PR #24, opened 2018-04-09
Add configuration files

mvsusp:
# Add configuration files

I'm closing https://github.com/aws/sagemaker-container-support/pull/23 since both have the setup.py file changes. 

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,39 @@
+import os
+from glob import glob
+from os.path import basename
+from os.path import splitext
+
+from setuptools import find_packages, setup
+
+
+def read(file_name):
+    return open(os.path.join(os.path.dirname(__file__), file_name)).read()
+
+
+
+setup(
+    name='sagemaker-containers',
+    version='1.0',
+    description='Open source library for creating containers to run on Amazon SageMaker.',
+
+    packages=find_packages(where='src', exclude=('test',)),
+    package_dir={'': 'src'},
+    py_modules=[splitext(basename(path))[0] for path in glob('src/*.py')],
+    long_description=read('README.md'),
+    author='Amazon Web Services',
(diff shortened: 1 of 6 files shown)


laurenyu:
can you fix the build error?

tox.ini
```diff
@@ -0,0 +1,53 @@
...
+    teamcity-messages
+    awslogs
+
+[testenv:flake8]
+basepython = python
+deps =
+    flake8
+    pep8-naming
+    flake8-import-order
+commands = flake8
```
laurenyu: don't forget a newline at the end of the file

mvsusp: done

setup.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def read(file_name):
+    return open(os.path.join(os.path.dirname(__file__), file_name)).read()
+
+
+
+setup(
+    name='sagemaker-container-support',
+    version='1.0',
```
winstonaws: Version should be 2.0.0

setup.py
```diff
@@ -0,0 +1,39 @@
...
+from setuptools import find_packages, setup
+
+
+def read(file_name):
+    return open(os.path.join(os.path.dirname(__file__), file_name)).read()
+
+
+
+setup(
+    name='sagemaker-container-support',
```
winstonaws: Name should match our existing name: https://github.com/aws/sagemaker-container-support/blob/master/setup.py#L16

so sagemaker_container_support

I know it doesn't follow python convention, but I think if we're keeping the same name, then continuity is important. (If we decide to change the name, everything changes, of course.)

mvsusp:
@laurenyu @nadiaya , Yuan will work in the name. Using the old name in this PR.

PR #27, opened 2018-04-15
Mvs create environment

mvsusp:
# Environment calls and environment functions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,6 @@
 .pytest_cache/
 .tox/
 src/sagemaker_containers.egg-info/
+htmlcov
+**/sagemaker_container_support.egg-info
+**/*.pyc
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ src/sagemaker_containers.egg-info/
 htmlcov
 **/sagemaker_container_support.egg-info
 **/*.pyc
+mypy_cache**
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,6 @@ def read(file_name):
     return open(os.path.join(os.path.dirname(__file__), file_name)).read()
 
 
-
 setup(
     name='sagemaker_container_support',
     version='2.0.0',
@@ -34,6 +33,6 @@ setup(
         'Programming Language :: Python :: 3.5',
     ],
     extras_require={
-        'test': ['tox', 'flake8', 'pytest', 'pytest-cov']
-    }
+        'test': ['tox', 'flake8', 'pytest', 'pytest-cov', 'mock']
+    }, install_requires=['boto3', 'PyYAML', 'six']
 )
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,9 @@ setup(
         'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3.5',
     ],
+    install_requires=['boto3', 'six', 'typing'],
+
     extras_require={
         'test': ['tox', 'flake8', 'pytest', 'pytest-cov', 'mock']
-    }, install_requires=['boto3', 'PyYAML', 'six']
+    }
 )
--- a/src/sagemaker_container_support/environment.py
+++ b/src/sagemaker_container_support/environment.py
@@ -18,6 +18,7 @@ import multiprocessing
 import os
 import shlex
 import subprocess
+from typing import Any, Dict, List, Tuple, Union  # noqa: F401
 
 import boto3
 
@@ -31,59 +32,57 @@ else:
 logging.basicConfig()
 logger = logging.getLogger(__name__)
 
-BASE_PATH_ENV = 'BASE_PATH'
-CURRENT_HOST_ENV = 'CURRENT_HOST'
-JOB_NAME_ENV = 'JOB_NAME'
-USE_NGINX_ENV = 'SAGEMAKER_USE_NGINX'
+BASE_PATH_ENV = 'BASE_PATH'  # type: str
+CURRENT_HOST_ENV = 'CURRENT_HOST'  # type: str
+JOB_NAME_ENV = 'JOB_NAME'  # type: str
+USE_NGINX_ENV = 'SAGEMAKER_USE_NGINX'  # type: str
 
-BASE_PATH = os.environ.get(BASE_PATH_ENV, '/opt/ml')
+BASE_PATH = os.environ.get(BASE_PATH_ENV, os.path.join('/opt', 'ml'))  # type: str
 
-MODEL_PATH = os.path.join(BASE_PATH, 'model')
-INPUT_PATH = os.path.join(BASE_PATH, 'input')
-INPUT_DATA_PATH = os.path.join(INPUT_PATH, 'data')
-INPUT_CONFIG_PATH = os.path.join(INPUT_PATH, 'config')
-OUTPUT_PATH = os.path.join(BASE_PATH, 'output')
-OUTPUT_DATA_PATH = os.path.join(OUTPUT_PATH, 'data')
+MODEL_PATH = os.path.join(BASE_PATH, 'model')  # type: str
+INPUT_PATH = os.path.join(BASE_PATH, 'input')  # type: str
+INPUT_DATA_PATH = os.path.join(INPUT_PATH, 'data')  # type: str
+INPUT_CONFIG_PATH = os.path.join(INPUT_PATH, 'config')  # type: str
+OUTPUT_PATH = os.path.join(BASE_PATH, 'output')  # type: str
+OUTPUT_DATA_PATH = os.path.join(OUTPUT_PATH, 'data')  # type: str
 
-HYPERPARAMETERS_FILE = 'hyperparameters.json'
-RESOURCE_CONFIG_FILE = 'resourceconfig.json'
-INPUT_DATA_CONFIG_FILE = 'inputdataconfig.json'
+HYPERPARAMETERS_FILE = 'hyperparameters.json'  # type: str
+RESOURCE_CONFIG_FILE = 'resourceconfig.json'  # type: str
+INPUT_DATA_CONFIG_FILE = 'inputdataconfig.json'  # type: str
 
-PROGRAM_PARAM = 'sagemaker_program'
-SUBMIT_DIR_PARAM = 'sagemaker_submit_directory'
-ENABLE_METRICS_PARAM = 'sagemaker_enable_cloudwatch_metrics'
-LOG_LEVEL_PARAM = 'sagemaker_container_log_level'
-JOB_NAME_PARAM = 'sagemaker_job_name'
-DEFAULT_MODULE_NAME_PARAM = 'default_user_module_name'
-REGION_PARAM_NAME = 'sagemaker_region'
+PROGRAM_PARAM = 'sagemaker_program'  # type: str
+SUBMIT_DIR_PARAM = 'sagemaker_submit_directory'  # type: str
+ENABLE_METRICS_PARAM = 'sagemaker_enable_cloudwatch_metrics'  # type: str
+LOG_LEVEL_PARAM = 'sagemaker_container_log_level'  # type: str
+JOB_NAME_PARAM = 'sagemaker_job_name'  # type: str
+DEFAULT_MODULE_NAME_PARAM = 'default_user_module_name'  # type: str
+REGION_PARAM_NAME = 'sagemaker_region'  # type: str
 
 SAGEMAKER_HPS = [PROGRAM_PARAM, SUBMIT_DIR_PARAM, ENABLE_METRICS_PARAM, REGION_PARAM_NAME,
-                 LOG_LEVEL_PARAM, JOB_NAME_PARAM, DEFAULT_MODULE_NAME_PARAM]
+                 LOG_LEVEL_PARAM, JOB_NAME_PARAM, DEFAULT_MODULE_NAME_PARAM]  # type: List[str]
 
 
-def read_json(path):
+def read_json(path):  # type: (str) -> Dict[str, Any]
     """Read a JSON file.
 
     Args:
-        path (string): path to the file
+        path (str): Path to the file.
 
     Returns:
-        A Python dictionary representation of the JSON file.
+        (dict[object, object]): A dictionary representation of the JSON file.
     """
     with open(path, 'r') as f:
         return json.load(f)
 
 
-def read_hyperparameters():
+def read_hyperparameters():  # type: () -> Dict[str, object]
     """Read the hyperparameters from /opt/ml/input/config/hyperparameters.json.
-
 
     For more information about hyperparameters.json:
     https://docs.aws.amazon.com/sagemaker/latest/dg/your-algorithms-training-algo.html#your-algorithms-training-algo-running-container-hyperparameters
 
     Returns:
-         hyperparameters (dict[string, object]): a map containing the hyperparameters.
-
+         (dict[string, object]): a dictionary containing the hyperparameters.
     """
     hyperparameters = read_json(os.path.join(INPUT_CONFIG_PATH, HYPERPARAMETERS_FILE))
 
@@ -99,13 +98,15 @@ def read_hyperparameters():
         reraise(e)
 
 
-def split_hyperparameters(hyperparameters, keys=SAGEMAKER_HPS):
+def split_hyperparameters(hyperparameters,  # type: Dict[str, object]
+                          keys=SAGEMAKER_HPS  # type: List[str]
+                          ):  # type: (...) -> Tuple[Dict[str, object], Dict[str, object]]
     """Split a dictionary in two by the provided keys. The default key SAGEMAKER_HPS splits user provided
     hyperparameters from SageMaker Python SDK provided hyperparameters.
 
     Args:
-        hyperparameters: A Python dictionary
-        keys: Lists of keys which will be the split criteria
+        hyperparameters (dict[str, object]): A Python dictionary
+        keys (list[str]): Lists of keys which will be the split criteria
 
     Returns:
         criteria (dict[string, object]), not_criteria (dict[string, object]): the result of the split criteria.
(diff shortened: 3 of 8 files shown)


setup.py
```diff
@@ -33,7 +32,9 @@ def read(file_name):
         'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3.5',
     ],
+    install_requires=['boto3', 'six', 'typing'],
```
jesterhazy: we should not include typing. according to the docs, it's provisional, and may see breaking changes even in minor releases.

tox.ini
```diff
@@ -43,7 +44,7 @@ deps =
     contextlib2
     teamcity-messages
     awslogs
-
+    typing
```
jesterhazy: see prev comment about typing

src/sagemaker_container_support/environment.py
```diff
@@ -0,0 +1,607 @@
...
+import boto3
+
+from six import PY2, reraise
+
+if PY2:
+    JSONDecodeError = None
+else:
+    from json.decoder import JSONDecodeError
+
+logging.basicConfig()
```
jesterhazy: can we remove this now? we should only call basicConfg at app entrypoints/main functions

src/sagemaker_container_support/environment.py
```diff
@@ -0,0 +1,607 @@
...
+
+PROGRAM_PARAM = 'sagemaker_program'  # type: str
+SUBMIT_DIR_PARAM = 'sagemaker_submit_directory'  # type: str
+ENABLE_METRICS_PARAM = 'sagemaker_enable_cloudwatch_metrics'  # type: str
+LOG_LEVEL_PARAM = 'sagemaker_container_log_level'  # type: str
+JOB_NAME_PARAM = 'sagemaker_job_name'  # type: str
+DEFAULT_MODULE_NAME_PARAM = 'default_user_module_name'  # type: str
+REGION_PARAM_NAME = 'sagemaker_region'  # type: str
+
+SAGEMAKER_HPS = [PROGRAM_PARAM, SUBMIT_DIR_PARAM, ENABLE_METRICS_PARAM, REGION_PARAM_NAME,
```
jesterhazy: _HPS -> _HYPERPARAMETERS

src/sagemaker_container_support/environment.py
```diff
@@ -0,0 +1,607 @@
...
+
+if PY2:
+    JSONDecodeError = None
+else:
+    from json.decoder import JSONDecodeError
+
+logging.basicConfig()
+logger = logging.getLogger(__name__)
+
+BASE_PATH_ENV = 'BASE_PATH'  # type: str
```
jesterhazy: remove type hints

src/sagemaker_container_support/environment.py
```diff
@@ -0,0 +1,607 @@
...
+
+def gpu_count():  # type: () -> int
+    """The number of gpus available in the current container.
+
+    Returns:
+        (int): number of gpus available in the current container.
+    """
+    try:
+        cmd = shlex.split('nvidia-smi --list-gpus')
+        output = str(subprocess.check_output(cmd))
```
jesterhazy: returns "b'...'" in py3 (need to .decode('utf8'))

src/sagemaker_container_support/environment.py
```diff
@@ -0,0 +1,607 @@
...
+                                saved.
+
+            current_host (str): The name of the current container on the container network. For example, 'algo-1'.
+
+            num_gpu (int): The number of gpus available in the current container.
+
+            num_cpu (int): The number of cpus available in the current container.
+
+        Returns:
+            A `Environment` object.
```
jesterhazy: An

src/sagemaker_container_support/environment.py
```diff
@@ -0,0 +1,607 @@
...
+                   hosts=hosts,
+                   channel_input_dirs={channel: channel_path(channel) for channel in input_data_config},
+                   num_gpu=gpu_count(),
+                   num_cpu=cpu_count(),
+                   hyperparameters=hyperparameters,
+                   resource_config=resource_config,
+                   input_data_config=read_input_data_config(),
+                   module_name=str(sagemaker_hyperparameters.get(PROGRAM_PARAM)),
+                   module_dir=str(sagemaker_hyperparameters.get(SUBMIT_DIR_PARAM)),
+                   enable_metrics=sagemaker_hyperparameters.get(ENABLE_METRICS_PARAM, False),
```
jesterhazy: what's the status of the in-container metrics? should we deprecate/remove them?

src/sagemaker_container_support/environment.py
mvsusp: Will do!

src/sagemaker_container_support/environment.py
mvsusp: I will add a test for it

src/sagemaker_container_support/environment.py
mvsusp: Maybe. Let’s tall about this offline.

src/sagemaker_container_support/environment.py
mvsusp: @jesterhazy  At this moment, this is the only module/file in the project so that setting up the config there is correct. 

Are you ok with us moving the logging to a different place in the moment that more modules need it?

jesterhazy: this is a library consumed by other projects... even more reason not to configure logging. move it to tests?

src/sagemaker_container_support/environment.py
jesterhazy: agree, pulling that is off-topic for this PR

test/unit/test_environment.py
```diff
@@ -0,0 +1,160 @@
...
+    assert env.channel_input_dirs['validation'] == '/opt/ml/input/data/validation'
+    assert env.current_host == RESOURCE_CONFIG['current_host']
+    assert env.module_name == 'main.py'
+    assert env.module_dir == 'imagenet'
+    assert env.enable_metrics
+    assert env.log_level == logging.WARNING
+
+
+def write_json(name, data, file_path):
+    file_path = file_path.join(name)
```
iquintero: I know this is test code but we should try to have the same bar too. Why can't this function be 

```python
write_json_file(file, data):
```

no need to do that join inside this function. Also no need to return anything if you are already receiving the full path to the file as an argument.

Also, if you are doing a join it should be os.path.join()  to make sure this works on windows too.

test/unit/test_environment.py
```diff
@@ -0,0 +1,160 @@
...
+    assert environment.channel_path('training') == str(input_data_path.join('training'))
+
+
+@patch('subprocess.check_output', lambda s: b('GPU 0\nGPU 1'))
+def test_gpu_count_in_gpu_instance():
+    assert environment.gpu_count() == 2
+
+
+def test_gpu_count_in_cpu_instance():
+    assert environment.gpu_count() == 0
```
iquintero: if I run this in a box with an nvidia gpu + nvidia-smi installed it will fail. This is not that uncommon.

setup.py
mvsusp: @jesterhazy, the typing module is only used here for type hinting. Type hinting was introduced by Guido himself in python 3.5. It does not enforce code. I used type hinting here with the objective to be more informative about the functions signatures to our users. The only thing it really does is: 

screen shot 2018-04-16 at 3 55 16 am
screen shot 2018-04-16 at 3 57 00 am

You will notice in the codebase that typing is only imported once, and it has being used in the typing comments. 

I agree that we should think about using typing out of the scope of this PR. I will remove it.

test/unit/test_environment.py
```diff
@@ -0,0 +1,160 @@
...
+
+    assert environment.read_input_data_config() == INPUT_DATA_CONFIG
+
+
+def test_channel_input_dirs(input_data_path):
+    assert environment.channel_path('evaluation') == str(input_data_path.join('evaluation'))
+    assert environment.channel_path('training') == str(input_data_path.join('training'))
+
+
+@patch('subprocess.check_output', lambda s: b('GPU 0\nGPU 1'))
```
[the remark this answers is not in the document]
mvsusp: https://sagemaker-build.corp.amazon.com/viewLog.html?buildId=5980&buildTypeId=ContainerSupport_Public_PullRequests&tab=buildLog&_focus=260#_state=156

src/sagemaker_container_support/environment.py
```diff
@@ -0,0 +1,607 @@
...
+    For more information about hyperparameters.json:
+    https://docs.aws.amazon.com/sagemaker/latest/dg/your-algorithms-training-algo.html#your-algorithms-training-algo-running-container-hyperparameters
+
+    Returns:
+         (dict[string, object]): a dictionary containing the hyperparameters.
+    """
+    hyperparameters = read_json(os.path.join(INPUT_CONFIG_PATH, HYPERPARAMETERS_FILE))
+
+    try:
+        return {k: json.loads(v) for k, v in hyperparameters.items()}
```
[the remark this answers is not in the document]
mvsusp: Yes, unfortunately we do: https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/estimator.py#L643

src/sagemaker_container_support/environment.py
```diff
@@ -0,0 +1,607 @@
...
+                                    'RecordWrapperType': 'None'
+                                },
+                                'validation': {
+                                    'TrainingInputMode': 'File',
+                                    'S3DistributionType': 'FullyReplicated',
+                                    'RecordWrapperType': 'None'
+                                }}
+
+                                You can find more information about /opt/ml/input/config/inputdataconfig.json here:
+                                https://docs.aws.amazon.com/sagemaker/latest/dg/your-algorithms-training-algo.html#your-algorithms-training-algo-running-container-inputdataconfig
```
andremoeller: I think you should move this link / comment into the docstring above Returns (and replace the link to https://docs.aws.amazon.com/sagemaker/latest/dg/your-algorithms-training-algo.html#your-algorithms-training-algo-running-container-dist-training with this link)

src/sagemaker_container_support/environment.py
```diff
@@ -0,0 +1,607 @@
...
+
+    For more information about channels: https://docs.aws.amazon.com/sagemaker/latest/dg/API_Channel.html
+
+    Returns:
+        (str) The input data directory for the specified channel.
+    """
+    return os.path.join(INPUT_DATA_PATH, channel)
+
+
+def gpu_count():  # type: () -> int
```
andremoeller: gpu_count -> num_gpus? Same as in train's function signature.

src/sagemaker_container_support/environment.py
```diff
@@ -0,0 +1,607 @@
...
+    try:
+        cmd = shlex.split('nvidia-smi --list-gpus')
+        output = subprocess.check_output(cmd).decode('utf-8')
+        return sum([1 for x in output.split('\n') if x.startswith('GPU ')])
+    except (OSError, subprocess.CalledProcessError):
+        logger.warning('No GPUs detected (normal if no gpus installed)')
+        return 0
+
+
+def cpu_count():  # type: () -> int
```
andremoeller: cpu_count -> num_cpus? Same as in train's function signature.

src/sagemaker_container_support/environment.py
```diff
@@ -0,0 +1,607 @@
...
+        # get the folder where the model should be saved
+        >>>model_dir = env.model_dir
+
+        >>>data = np.load(os.path.join(training_dir, training_data_file))
+
+        >>>x_train, y_train = data['features'], keras.utils.to_categorical(data['labels'])
+
+        >>>model = ResNet50(weights='imagenet')
+
+        unfreeze the model to allow fine tuning
```
andremoeller: "unfreeze the model to allow fine tuning" what does this mean? and what are ... for (if this is a TODO, can you add "TODO"?

.gitignore
```diff
@@ -5,3 +5,7 @@
 .pytest_cache/
 .tox/
 src/sagemaker_containers.egg-info/
```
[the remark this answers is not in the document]
mvsusp: deleted.

setup.py
mvsusp: @jesterhazy I removed the import from typing. I am still using the type hints comments convention, just with basic python types.

That changes the example above to:


screen shot 2018-04-19 at 6 56 54 am

Please, let me know if you are ok with that.

src/sagemaker_container_support/environment.py
mvsusp: Great point. I removed the logger config. I added a TODO to test how easy will be to other projects to configure logging in this app.

src/sagemaker_container_support/environment.py
mvsusp: The specific example above is just a comment without important any library type or making any enforcement.

Let me know if you are of with these kind of comment. I am not importing typing anymore.

src/sagemaker_container_support/environment.py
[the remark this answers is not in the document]
mvsusp: @laurenyu is redesigning how we pass the hyperparameters from the client to the container.

src/sagemaker_container_support/environment.py
```diff
@@ -0,0 +1,607 @@
...
+
+    Returns:
+         (dict[string, object]): a dictionary containing the hyperparameters.
+    """
+    hyperparameters = read_json(os.path.join(INPUT_CONFIG_PATH, HYPERPARAMETERS_FILE))
+
+    try:
+        return {k: json.loads(v) for k, v in hyperparameters.items()}
+    except (JSONDecodeError, TypeError):
+        return hyperparameters
```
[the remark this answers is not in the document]
mvsusp: Forgot to add the warning, thanks!

The first except happens in python 3 and the second except happens with python 2.

I have tests convering the cases:
     https://github.com/mvsusp/sagemaker-container-support/blob/a5b06d1d76d284adc7cd964867fc1d90fe651669/test/unit/test_environment.py#L73

src/sagemaker_container_support/environment.py
```diff
@@ -0,0 +1,607 @@
...
+    Args:
+        hyperparameters (dict[str, object]): A Python dictionary
+        keys (list[str]): Lists of keys which will be the split criteria
+
+    Returns:
+        criteria (dict[string, object]), not_criteria (dict[string, object]): the result of the split criteria.
+    """
+    not_keys = set(hyperparameters.keys()) - set(keys)
+
+    return {k: hyperparameters[k] for k in keys if k in hyperparameters}, {k: hyperparameters[k] for k in not_keys}
```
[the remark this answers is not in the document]
mvsusp: refactored

src/sagemaker_container_support/environment.py
mvsusp: Agreed! Thank you.

test/unit/test_environment.py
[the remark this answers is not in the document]
mvsusp: Your test dummy does not work because @patch requires either a second attribute or an extra argument to the decorated functions.

From the docs: https://github.com/testing-cabal/mock/blob/master/mock/mock.py#L1605

You can notice that all the testes are providing a lambda function that is the new object.

Tested using tox:
screen shot 2018-04-19 at 7 33 18 am

I am configuring teamcity today to have the same nice experience.

test/unit/test_environment.py
mvsusp: Good point, patching it as well.

test/unit/test_environment.py
mvsusp: Good point:

What about ```json_dump```? I believe is more pythonic than ```write_json_file```. 

All the naming in this project so far was strong based on python standard lib  naming conventions and glossary: https://docs.python.org/3/library/ and https://docs.python.org/3/glossary.html

Naming is hard, so let's sync up offline to see if this strategy is good for us.

file_path is not a string but a py.path.local. Which is pytest default way to create temp files: https://docs.pytest.org/en/latest/tmpdir.html

I refactored the functions to the following:
```python
def json_dump(data, path_obj):  # type: (object, py.path.local) -> None
    """Writes JSON serialized data to the local file system path

    Args:
        data (object): object to be serialized
        path_obj (py.path.local): path.local object of the file to be written
    """
    path_obj.write(json.dumps(data))
```

And the tests look like this now:
```python
def test_read_hyperparameters(input_config_path):
    json_dump(ALL_HPS, input_config_path.join('hyperparameters.json'))

    assert environment.read_hyperparameters() == ALL_HPS
```

Let me know your thoughts.

test/unit/test_environment.py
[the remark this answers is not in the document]
jesterhazy: i'm also surprised this works, but it does and makes waaay more sense than the usual patch usage!!

tox.ini
```diff
@@ -33,8 +33,9 @@ passenv =
 # {posargs} can be passed in by additional arguments specified when invoking tox.
 # Can be used to specify which tests to run, e.g.: tox -- -s
 commands =
-    coverage run --source sagemaker_containers -m py.test {posargs}
+    coverage run --source sagemaker_container_support -m py.test {posargs}
```
jesterhazy: we need to fix project name, package name, and source code layout so they are aligned. since the repo is called `sagemaker-containers`, the package name should be `sagemaker_containers` and the directory structure should be `src/sagemaker_containers`

but **important** do that in separate PR to avoid the features / file moves combo.

src/sagemaker_container_support/environment.py
mvsusp: Because of design decisions. @winstonaws and I came with the following design:

```Environment``` will be used by our customers. It is just a class with getters and setters well documented explaining how sagemaker environment works.

The entire framework is made by composition of small functions. These small functions allows users adhere to high level solution or just use the small functions to extend their needs.

The entire framework was refactored from the bottom up to that objective.

File: src/sagemaker_container_support/environment.py [hunk 21]
```diff
@@ -0,0 +1,600 @@
...
+import boto3
+
+from six import PY2, reraise
+
+if PY2:
+    JSONDecodeError = None
+else:
+    from json.decoder import JSONDecodeError
+
+# TODO (mvsusp): tests for how other apps will set different logger levels
```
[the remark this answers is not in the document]
mvsusp: @nadiaya @andreataws and me are adding TODOs (same approach as tensorflow) with our github aliases on it. I believe is a very informative way to communicate.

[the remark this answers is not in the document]
mvsusp: Agreed. I was thinking more specifically in cases like boto and botocore that are super noise and need to be handled differently. Like you did before.

tox.ini
mvsusp: I have another branch with the name changes waiting for this one

PR #33, opened 2018-04-20
Import download and install modules

mvsusp:
Add functionality to deal with modules.


test/functional/conftest.py
```diff
@@ -0,0 +1,67 @@
...
+import os
+import time
+
+import boto3
+
+import pytest
+
+from sagemaker import Session, fw_utils
+
+import six
```
andremoeller: I think six should be with boto3

src/sagemaker_containers/modules.py
```diff
@@ -0,0 +1,80 @@
...
+import shutil
+import subprocess
+import sys
+import tarfile
+import tempfile
+import traceback
+
+import boto3
+
+from six.moves.urllib.parse import urlparse
```
andremoeller: I think this import should be with boto3

src/sagemaker_containers/modules.py
```diff
@@ -0,0 +1,80 @@
...
+                     'setup(name="%s", py_modules=["%s"])' % (name, name)]
+
+            f.write(os.linesep.join(lines))
+
+
+def install(path):
+    try:
+        subprocess.check_call(shlex.split('%s -m pip install %s -U' % (sys.executable, path)))
+    except subprocess.CalledProcessError:
+        raise ValueError('Failed to pip install %s:%s%s' % (path, os.linesep, traceback.format_exc()))
```
andremoeller: I don't think this is a ValueError. Maybe just RuntimeError?

test/unit/test_modules.py
```diff
@@ -0,0 +1,127 @@
...
+# language governing permissions and limitations under the License.
+from __future__ import absolute_import
+
+import os
+import subprocess
+import sys
+
+from mock import call, mock_open, patch
+
+import pytest
```
andremoeller: Same as other comments: organize imports

src/sagemaker_containers/modules.py
mvsusp: Sounds good.

src/sagemaker_containers/__init__.py
```diff
@@ -12,6 +12,7 @@
 # language governing permissions and limitations under the License.
 from __future__ import absolute_import
 
+from sagemaker_containers import modules
```
winstonaws: Nitpick - does importing modules here change anything? A user would still refer to it as sagemaker_containers.modules, right?

And for adding it to __all__, my understanding is that this controls the "from sagemaker_containers import *" behavior. In this case, we don't really want "modules" to be imported into the current user's namespace as is, do we?

src/sagemaker_containers/modules.py
```diff
@@ -0,0 +1,79 @@
...
+
+import boto3
+from six.moves.urllib.parse import urlparse
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_MODULE_NAME = 'default_user_module_name'
+
+
+def download(url, dst):
```
winstonaws: Maybe "s3_download" since this is s3 specific?

src/sagemaker_containers/modules.py
```diff
@@ -0,0 +1,79 @@
...
+    s3 = boto3.resource('s3')
+    s3.Bucket(bucket).download_file(key, dst)
+
+
+def prepare(path, name):
+    if not os.path.exists(os.path.join(path, 'setup.py')):
+        logging.info('Module %s does not provide a setup.py. Generating a minimal setup.' % name)
+
+        with open(os.path.join(path, 'setup.py'), 'w') as f:
+            lines = ['from distutils.core import setup',
```
winstonaws: Should we use setuptools instead of distutils? https://packaging.python.org/guides/tool-recommendations/

src/sagemaker_containers/modules.py
```diff
@@ -0,0 +1,79 @@
...
+                     'setup(name="%s", py_modules=["%s"])' % (name, name)]
+
+            f.write(os.linesep.join(lines))
+
+
+def install(path):
+    try:
+        subprocess.check_call(shlex.split('%s -m pip install %s -U' % (sys.executable, path)))
```
winstonaws: Just wondering - will this do the right thing when running inside a conda env / virtualenv?

src/sagemaker_containers/modules.py
```diff
@@ -0,0 +1,79 @@
...
+                tmpdir = tempfile.mkdtemp()
+                try:
+                    t.extractall(path=tmpdir)
+
+                    prepare(tmpdir, name)
+
+                    install(tmpdir)
+                finally:
+                    shutil.rmtree(tmpdir)
+                    return importlib.import_module(name)
```
winstonaws: This return should go inside the try, no?

setup.py
```diff
@@ -32,9 +30,9 @@ def read(file_name):
         'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3.5',
     ],
-    install_requires=['boto3', 'six', 'typing'],
+    install_requires=['boto3', 'six', 'pip'],
 
     extras_require={
-        'test': ['tox', 'flake8', 'pytest', 'pytest-cov', 'mock']
+        'test': ['tox', 'flake8', 'pytest', 'pytest-cov', 'mock', 'sagemaker']
```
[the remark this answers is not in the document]
mvsusp: To import the session is fw_utils: https://github.com/aws/sagemaker-containers/pull/33/files#diff-cb4f71ec7457ece91d85d328b489f3daR19

src/sagemaker_containers/__init__.py
mvsusp: It is actually not a nitpick @winstonaws. We need to keep modules so we can import like this:
```python 
import sagemaker-containers as smc

smc.modules.something
```

You are correct that ```__all__``` imports everything. I am removing it.

src/sagemaker_containers/modules.py
mvsusp: sure.

mvsusp: Actually, my thought process is the following:
download for now is only S3 and throws exception otherwise.
later, if we are downloading for git for example, we can rename download to s3_download, create a clone function and create a download functions that cover both cases. What do you think about this approach?

src/sagemaker_containers/modules.py
mvsusp: Good call.

src/sagemaker_containers/modules.py
mvsusp: Teamcity is using conda, I don't think anything different would happen in env/virtualenv. In the works case scenario, ```sys.executable``` returns ```None``` or ```''```.  I will add unit test for these cases.

src/sagemaker_containers/modules.py
mvsusp: Yeah!

src/sagemaker_containers/modules.py
winstonaws: I think even for now, with only having one function, we should give it a name which is descriptive enough. I get that in general we want short names since its pythonic, but I think the three characters here are worth it.

test/unit/test_modules.py
```diff
@@ -0,0 +1,132 @@
...
+@patch('boto3.resource', autospec=True)
+@pytest.mark.parametrize('url,bucket_name,key,dst', [
+    ('S3://my-bucket/path/to/my-file', 'my-bucket', 'path/to/my-file', '/tmp/my-file'),
+    ('s3://my-bucket/my-file', 'my-bucket', 'my-file', '/tmp/my-file')
+])
+def test_s3_download(resource, url, bucket_name, key, dst):
+    smc.modules.s3_download(url, dst)
+
+    chain = call('s3').Bucket(bucket_name).download_file(key, dst)
+    assert resource.mock_calls == chain.call_list()
```
winstonaws: Does this test correctly verify the arguments passed to .Bucket and download_file? e.g. if you change the parameters in the call to create "chain", does the test fail?

test/functional/test_download_and_import.py
```diff
@@ -0,0 +1,40 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
mvsusp: Sure.

test/functional/test_download_and_import.py
```diff
@@ -0,0 +1,40 @@
...
+# distributed on an 'AS IS' BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from __future__ import absolute_import
+
+from sagemaker_containers import modules
+
+
+def test_download_and_import_module(upload_script, create_script):
+    create_script('my-script.py', 'def validate(): return True')
```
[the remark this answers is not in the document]
mvsusp: Sure

test/functional/test_download_and_import.py
```diff
@@ -0,0 +1,40 @@
...
+
+    content = ['from distutils.core import setup',
```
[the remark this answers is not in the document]
mvsusp: Agreed

test/unit/test_modules.py
mvsusp: It does ! https://docs.python.org/3/library/unittest.mock-examples.html#setting-return-values-and-attributes

src/sagemaker_containers/modules.py
```diff
@@ -0,0 +1,115 @@
...
+import boto3
+from six.moves.urllib.parse import urlparse
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_MODULE_NAME = 'default_user_module_name'
+
+
+def s3_download(url, dst):  # type: (str, str) -> None
+    """Download a file from the S3.
```
andremoeller: "the S3" -> "S3"

src/sagemaker_containers/modules.py
```diff
@@ -0,0 +1,115 @@
...
+
+    Args:
+        url (str): the s3 url location of the file.
```
andremoeller: "url location" -> "url"

src/sagemaker_containers/modules.py
```diff
@@ -0,0 +1,115 @@
...
+    if not sys.executable:
+        raise RuntimeError('Failed to retrieve the real path for the Python executable binary')
+    try:
+        subprocess.check_call(shlex.split('%s -m pip install %s -U' % (sys.executable, path)))
+    except subprocess.CalledProcessError:
+        raise RuntimeError('Failed to pip install %s:%s%s' % (path, os.linesep, traceback.format_exc()))
+
+
+def download_and_import(url, name=DEFAULT_MODULE_NAME):  # type: (str, str) -> module
+    """Download, Prepare and Install a compressed tar file from S3 as a module.
```
andremoeller: "Download, Prepare and Install" -> "Download, prepare, and install"

src/sagemaker_containers/modules.py
```diff
@@ -0,0 +1,115 @@
...
+
+    SageMaker Python SDK saves the user provided scripts as compressed tar files in S3
+    https://github.com/aws/sagemaker-python-sdk#mxnet-sagemaker-estimators.
```
andremoeller: suggestion: "https://github.com/aws/sagemaker-python-sdk#mxnet-sagemaker-estimators" -> "https://github.com/aws/sagemaker-python-sdk" -- the former link will probably be broken at some point

src/sagemaker_containers/modules.py
```diff
@@ -0,0 +1,115 @@
...
+
+    This function downloads this compressed file, transforms it as a module, and installs it.
+
+    Args:
+        url (str): the s3 url location of the file.
+        name (str): name of the script or module.
+
+    Returns:
```
andremoeller: "Returns:" is blank.

PR #35, opened 2018-04-25
Environment extends map

mvsusp:
- Environment extend Mapping
- Improved Environment test coverage
- ignoring python 2 code coverage when running python 3 tests and vice versa

### DIFF
--- a/.coveragerc_py35
+++ b/.coveragerc_py35
@@ -1,0 +1,20 @@
+[run]
+branch = True
+timid = True
+
+[report]
+exclude_lines =
+    pragma: no cover
+    pragma: py3 no cover
+    if six.PY2
+    elif six.PY2
+
+partial_branches =
+    pragma: no cover
+    pragma: py3 no cover
+    if six.PY3
+    elif six.PY3
+
+show_missing = True
+
+fail_under = 90
--- a/.coveragerc_py35
+++ b/.coveragerc_py35
@@ -6,8 +6,8 @@ timid = True
 exclude_lines =
     pragma: no cover
     pragma: py3 no cover
-    if six.PY2
-    elif six.PY2
+    if six.PY3
+    elif six.PY3
 
 partial_branches =
     pragma: no cover
--- a/src/sagemaker_containers/collections.py
+++ b/src/sagemaker_containers/collections.py
@@ -1,0 +1,14 @@
+def split_by_criteria(dictionary, keys):  # type: (dict, set) -> (dict, dict)
+    """Split a dictionary in two by the provided keys.
+
+    Args:
+        dictionary (dict[str, object]): A Python dictionary
+        keys (set[str]): Set of keys which will be the split criteria
+
+    Returns:
+        criteria (dict[string, object]), not_criteria (dict[string, object]): the result of the split criteria.
+    """
+    dict_matching_criteria = {k: dictionary[k] for k in dictionary.keys() if k in keys}
+    dict_not_matching_criteria = {k: dictionary[k] for k in dictionary.keys() if k not in keys}
+
+    return dict_matching_criteria, dict_not_matching_criteria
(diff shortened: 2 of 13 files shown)


src/sagemaker_containers/environment.py
```diff
@@ -200,7 +187,7 @@ def cpu_count():  # type: () -> int
     return multiprocessing.cpu_count()
 
 
-class Environment(object):
+class Environment(collections.Mapping):
```
jesterhazy: what is rationale for this change?

src/sagemaker_containers/environment.py
```diff
@@ -568,7 +580,8 @@ def create(cls, session=None):  # type: (boto3.Session) -> Environment
 
         input_data_config = read_input_data_config()
 
-        sagemaker_hyperparameters, hyperparameters = split_hyperparameters(read_hyperparameters())
+        hps = read_hyperparameters()
```
jesterhazy: hyperparameters

src/sagemaker_containers/collections.py
```diff
@@ -0,0 +1,14 @@
+def split_by_criteria(dictionary, keys):  # type: (dict, set) -> (dict, dict)
```
jesterhazy: is this function used anywhere except Environment? if not, it should be private method in environment

mvsusp: Environment by design is a namedtuple, POCO, POJO, (put your plain class of choice here).
Another design decision is that this library is composed by small functions that can be utilized as well. Private methods and classes are only used when necessary.

I have changes outside the scope of this PR that already use split_by_criteria.

src/sagemaker_containers/environment.py
mvsusp: That name exists already https://github.com/mvsusp/sagemaker-containers/blob/597d9b0319e1c64b57a438df9417eca49e3e3e64/src/sagemaker_containers/environment.py#L420

jesterhazy: ya but we should take opportunity to fix

src/sagemaker_containers/environment.py
jesterhazy: discussed offline. ok.

src/sagemaker_containers/collections.py
jesterhazy: discussed offline ok.

src/sagemaker_containers/environment.py
```diff
@@ -34,7 +38,7 @@
 JOB_NAME_ENV = 'JOB_NAME'  # type: str
 USE_NGINX_ENV = 'SAGEMAKER_USE_NGINX'  # type: str
 
-BASE_PATH = os.environ.get(BASE_PATH_ENV, os.path.join('/opt', 'ml'))  # type: str
+BASE_PATH = os.environ.get(BASE_PATH_ENV, os.path.join(os.sep, 'opt', 'ml'))  # type: str
```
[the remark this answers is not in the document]
jesterhazy: if first arg to join has '/' it's ok, but if later ones do then it's trouble. looks like os.sep here is just to avoid platform-specific separator. but '/' should work fine on all platforms and is more readable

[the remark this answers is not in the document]
mvsusp: I will revert.

src/sagemaker_containers/environment.py
mvsusp: Should I move the Environment class to another file them? 
What do you suggest @jesterhazy

.coveragerc_py35
```diff
@@ -0,0 +1,20 @@
...
+exclude_lines =
+    pragma: no cover
+    pragma: py3 no cover
+    if six.PY2
+    elif six.PY2
+
+partial_branches =
+    pragma: no cover
+    pragma: py3 no cover
+    if six.PY3
```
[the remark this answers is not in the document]
mvsusp: Good catch!

PR #36, opened 2018-04-26
Mvs functions

mvsusp:
- add functions to match env attributes with function parameters

If you didn't have opportunity to review https://github.com/aws/sagemaker-containers/pull/35. 
I'm still handling comments even after closed.

### DIFF
--- a/.coveragerc_py27
+++ b/.coveragerc_py27
@@ -6,14 +6,14 @@ timid = True
 exclude_lines =
     pragma: no cover
     pragma: py2 no cover
-    if six.PY2
-    elif six.PY2
+    if six.PY3
+    elif six.PY3
 
 partial_branches =
     pragma: no cover
     pragma: py2 no cover
-    if six.PY2
-    elif six.PY2
+    if six.PY3
+    elif six.PY3
 
 show_missing = True
 
--- a/setup.py
+++ b/setup.py
@@ -33,6 +33,6 @@ setup(
     install_requires=['boto3', 'six', 'pip'],
 
     extras_require={
-        'test': ['tox', 'flake8', 'pytest', 'pytest-cov', 'mock', 'sagemaker']
+        'test': ['tox', 'flake8', 'pytest', 'pytest-cov', 'mock', 'sagemaker', 'keras', 'tensorflow', 'numpy']
     }
 )
--- a/src/sagemaker_containers/functions.py
+++ b/src/sagemaker_containers/functions.py
@@ -1,0 +1,78 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the 'License'). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the 'license' file accompanying this file. This file is
+# distributed on an 'AS IS' BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from __future__ import absolute_import
+
+import inspect
+
+import six
+
+import sagemaker_containers as smc
+
+
+def matching_args(fn, dictionary):  # type: (function, collections.Mapping) -> dict
+    """Given a function fn and a dict dictionary, returns the function arguments that match the dict keys.
+
+    Example:
+
+        def train(channel_dirs, model_dir): pass
+
+        dictionary = {'channel_dirs': {}, 'model_dir': '/opt/ml/model', 'other_args': None}
+
+        args = smc.functions.matching_args(train, dictionary) # {'channel_dirs': {}, 'model_dir': '/opt/ml/model'}
+
+        train(**args)
+    Args:
+        fn (function): a function
+        dictionary (dict): the dictionary with the keys
+
+    Returns:
+        (dict) a dictionary with only matching arguments.
+    """
+    args, _, kwargs = signature(fn)
+
+    if kwargs:
+        return dictionary
+
+    return smc.collections.split_by_criteria(dictionary, set(args))[0]
+
+
+def signature(fn):  # type: (function) -> ([], [], [])
+    """Given a function fn, returns the function args, vargs and kwargs
+
+    Args:
+        fn (function): a function
+
+    Returns:
+        ([], [], []): a tuple containing the function args, vargs and kwargs.
+    """
+    if six.PY2:
+        arg_spec = inspect.getargspec(fn)
+        return arg_spec.args, arg_spec.varargs, arg_spec.keywords
+    elif six.PY3:
+        sig = inspect.signature(fn)
+
+        def filter_parameters(kind):
+            return [
+                p.name for p in sig.parameters.values()
+                if p.kind == kind
+            ]
+
+        args = filter_parameters(inspect.Parameter.POSITIONAL_OR_KEYWORD)
+
+        vargs = filter_parameters(inspect.Parameter.VAR_POSITIONAL)
+
+        kwargs = filter_parameters(inspect.Parameter.VAR_KEYWORD)
+
+        return (args,
+                vargs[0] if vargs else None,
+                kwargs[0] if kwargs else None)
--- a/test/functional/test_keras_framework.py
+++ b/test/functional/test_keras_framework.py
@@ -1,0 +1,86 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from __future__ import absolute_import
+
+import logging
+import os
+
+import numpy as np
+
+import sagemaker_containers as smc
+
+dir_path = os.path.dirname(os.path.realpath(__file__))
+
+RESOURCE_CONFIG = dict(current_host='algo-1', hosts=['algo-1'])
+
+INPUT_DATA_CONFIG = {'training': {'ContentType': 'trainingContentType',
+                                  'TrainingInputMode': 'File',
+                                  'S3DistributionType': 'FullyReplicated',
+                                  'RecordWrapperType': 'None'}}
+
+hyperparameters = dict(training_data_file='training_data.npz', sagemaker_region='us-west-2',
+                       default_user_module_name='net',
+                       sagemaker_job_name='sagemaker-training-job', sagemaker_enable_cloudwatch_metrics=True,
+                       sagemaker_container_log_level=logging.WARNING, sagemaker_program='user_script.py')
+
+USER_SCRIPT = """
+import os
+
+import keras
+import numpy as np
+
+def train(channel_input_dirs, hyperparameters):
+    data = np.load(os.path.join(channel_input_dirs['training'], hyperparameters['training_data_file']))
+    x_train = data['features']
+    y_train = keras.utils.to_categorical(data['labels'], 10)
+
+    model = keras.models.Sequential()
+    model.add(keras.layers.Dense(10, activation='softmax', input_dim=1))
+
+    model.compile(loss='categorical_crossentropy', optimizer=keras.optimizers.SGD(), metrics=['accuracy'])
+
+    model.fit(x_train, y_train, epochs=1, batch_size=1)
+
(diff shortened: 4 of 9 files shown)


src/sagemaker_containers/functions.py
```diff
@@ -0,0 +1,78 @@
...
+
+import inspect
+
+import six
+
+import sagemaker_containers as smc
+
+
+def matching_args(fn, dictionary):  # type: (function, collections.Mapping) -> dict
+    """Given a function fn and a dict dictionary, returns the function arguments that match the dict keys.
```
andremoeller: I found this a bit confusing on first read-through. Maybe something like:

Given a function fn and a dict dictionary, returns a dictionary whose keys match (by name) the parameters of fn.

(I don't know if that's really better.)

src/sagemaker_containers/functions.py
```diff
@@ -0,0 +1,78 @@
...
+
+    if kwargs:
+        return dictionary
+
+    return smc.collections.split_by_criteria(dictionary, set(args))[0]
+
+
+def signature(fn):  # type: (function) -> ([], [], [])
+    """Given a function fn, returns the function args, vargs and kwargs
+
```
andremoeller: I think an example (with all 3: args, vargs, and kwargs) would be helpful here.

test/functional/test_keras_framework.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def test_keras_framework(create_channel, create_training):
+    create_training(script_name='user_script.py', script=USER_SCRIPT, hyperparameters=hyperparameters,
+                    resource_config=RESOURCE_CONFIG, input_data_config=INPUT_DATA_CONFIG)
+
+    features = np.random.random((10, 1))
+    labels = np.zeros((10, 1))
+
+    create_channel(channel='training', file_name='training_data.npz', data=dict(features=features, labels=labels))
```
andremoeller: Seems like `create_channel`, `create_training`, etc. are being removed (https://github.com/mvsusp/sagemaker-containers/pull/4/files)

Not necessary for this PR, but I'll be curious to see what these look like after.

src/sagemaker_containers/functions.py
```diff
@@ -0,0 +1,78 @@
...
+    Args:
+        fn (function): a function
+
+    Returns:
+        ([], [], []): a tuple containing the function args, vargs and kwargs.
```
winstonaws: I think we should return NamedTuples instead of regular tuples for all public functions unless we have a strong reason not to, since they still support the functionality of tuples, but gives the user a way to write more readable code if they want by referencing field names instead of tuple indices. An example is in the current matching_args implementation, which currently has:

smc.collections.split_by_criteria(dictionary, set(args))[0]

Which can become:

smc.collections.split_by_criteria(dictionary, set(args)).included (or whatever name we come up with for the field.)

You can see examples of this from the Python standard library, such as inspect.getargspec returning a ArgSpec named tuple.

If we agree on this, we should go back and apply this to the other places in our code where we return tuples (or revisit whether returning tuples at all is the best solution).

src/sagemaker_containers/functions.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+    if six.PY2:
+        arg_spec = inspect.getargspec(fn)
+        return arg_spec.args, arg_spec.varargs, arg_spec.keywords
```
winstonaws: Continuing with the above comment - I think we should just return the ArgSpec namedtuple as-is.

src/sagemaker_containers/functions.py
```diff
@@ -0,0 +1,78 @@
...
+    elif six.PY3:
+        sig = inspect.signature(fn)
```
winstonaws: I don't think we should use inspect.signature here if we're throwing away all of the extra information it gives us to return a subset of ArgSpec. We could either just use getargspec for both cases, or use getfullargspec and adapt the return type.

The docs for getfullargspec recommend this as well: https://docs.python.org/3/library/inspect.html#inspect.getfullargspec

"Changed in version 3.6: This method was previously documented as deprecated in favour of signature() in Python 3.5, but that decision has been reversed in order to restore a clearly supported standard interface for single-source Python 2/3 code migrating away from the legacy getargspec() API."

test/functional/test_keras_framework.py
mvsusp: That PR includes this branch, you can see the code here: https://github.com/mvsusp/sagemaker-containers/blob/924446d2325be7b5b634c9d4ee72e9fc3d49b6c1/test/functional/test_keras_framework.py#L72

src/sagemaker_containers/functions.py
mvsusp: @winstonaws great point. Will do that

src/sagemaker_containers/functions.py
mvsusp: Agreed.

src/sagemaker_containers/functions.py
mvsusp: Great point, I will do that.

src/sagemaker_containers/functions.py
mvsusp: i like it better as well. Thanks

src/sagemaker_containers/collections.py
```diff
@@ -10,19 +10,28 @@
 # distributed on an 'AS IS' BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+from __future__ import absolute_import
 
+import collections
 
-def split_by_criteria(dictionary, keys):  # type: (dict, set) -> (dict, dict)
+SplitByCriteriaSpec = collections.namedtuple('SplitByCriteriaSpec', 'included not_included')
```
winstonaws: How about "SplitResult" instead?

And "not_included" should be "excluded".

test/unit/test_functions.py
```diff
@@ -0,0 +1,40 @@
...
+    (lambda x, y='y': None, inspect.ArgSpec(['x', 'y'], None, None, ('y',))),
+    (lambda *args: None, inspect.ArgSpec([], 'args', None, None)),
+    (lambda **kwargs: None, inspect.ArgSpec([], None, 'kwargs', None)),
+    (lambda x, y, *args, **kwargs: None, inspect.ArgSpec(['x', 'y'], 'args', 'kwargs', None))
+])
+def test_signature(fn, expected):
+    assert smc.functions.getargspec(fn) == expected
+
+
+@pytest.mark.parametrize('fn, environment, expected', [
```
winstonaws: Let's add some cases to fully define the behavior of the function (even though they arent new branches):

* A fn which has both regular args and kwargs
* A fn which has regular args which aren't in the environment

src/sagemaker_containers/functions.py
```diff
@@ -0,0 +1,68 @@
...
+
+    Returns:
+        (dict) a dictionary with only matching arguments.
+    """
+    arg_spec = getargspec(fn)
+
+    if arg_spec.keywords:
+        return dictionary
+
+    return smc.collections.split_by_criteria(dictionary, set(arg_spec.args)).included
```
winstonaws: Is set() needed here? It's not possible for a function to have duplicate argument names, is it?

src/sagemaker_containers/collections.py
mvsusp: Agreed.

src/sagemaker_containers/functions.py
mvsusp: It could be a list but ```split_by_criteria``` only is receiving sets for now. I will change it to accept lists as well.

setup.py
```diff
@@ -33,6 +33,6 @@ def read(file_name):
     install_requires=['boto3', 'six', 'pip'],
 
     extras_require={
-        'test': ['tox', 'flake8', 'pytest', 'pytest-cov', 'mock', 'sagemaker']
+        'test': ['tox', 'flake8', 'pytest', 'pytest-cov', 'mock', 'sagemaker', 'keras', 'tensorflow', 'numpy']
```
nadiaya: shouldn't this package be agnostic of any frameworks?

test/functional/test_keras_framework.py
```diff
@@ -0,0 +1,86 @@
...
+
+import numpy as np
+
+import sagemaker_containers as smc
+
+dir_path = os.path.dirname(os.path.realpath(__file__))
+
+RESOURCE_CONFIG = dict(current_host='algo-1', hosts=['algo-1'])
+
+INPUT_DATA_CONFIG = {'training': {'ContentType': 'trainingContentType',
```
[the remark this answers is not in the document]
mvsusp: inputdataconfig.json is a default file from SageMaker training: https://github.com/aws/sagemaker-containers/blob/r2.0/src/sagemaker_containers/environment.py#L122

The difference of these setting can be find in the docstrings of the environment.py, aws docs (https://docs.aws.amazon.com/sagemaker/latest/dg/your-algorithms-training-algo.html#your-algorithms-training-algo-running-container-dist-training)

.coveragerc_py27
```diff
@@ -6,14 +6,14 @@ timid = True
 exclude_lines =
     pragma: no cover
     pragma: py2 no cover
-    if six.PY2
-    elif six.PY2
+    if six.PY3
```
nadiaya: should it really be 'six.PY3' in both places?
just curious: how does this logic works?

test/functional/test_keras_framework.py
```diff
@@ -0,0 +1,86 @@
...
+
+hyperparameters = dict(training_data_file='training_data.npz', sagemaker_region='us-west-2',
+                       default_user_module_name='net',
+                       sagemaker_job_name='sagemaker-training-job', sagemaker_enable_cloudwatch_metrics=True,
+                       sagemaker_container_log_level=logging.WARNING, sagemaker_program='user_script.py')
+
+USER_SCRIPT = """
+import os
+
+import keras
```
nadiaya: why do we need to test it for a specific framework?
I don't think we have or should have any keras functionality in the container.

test/functional/test_keras_framework.py
```diff
@@ -0,0 +1,86 @@
...
+dir_path = os.path.dirname(os.path.realpath(__file__))
+
+RESOURCE_CONFIG = dict(current_host='algo-1', hosts=['algo-1'])
+
+INPUT_DATA_CONFIG = {'training': {'ContentType': 'trainingContentType',
+                                  'TrainingInputMode': 'File',
+                                  'S3DistributionType': 'FullyReplicated',
+                                  'RecordWrapperType': 'None'}}
+
+hyperparameters = dict(training_data_file='training_data.npz', sagemaker_region='us-west-2',
```
nadiaya: shouldn't region be configurable or we are running this test purely locally?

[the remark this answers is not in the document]
nadiaya: But shouldn't it be capitalized for that exact reason?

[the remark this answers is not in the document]
mvsusp: It should! Good call. Renamed to HYPERPARAMETERS

.coveragerc_py27
mvsusp: it is a string match. Any code branch that starts with ```if six.PY3``` will be removed from python 2 coverage checking.

setup.py
mvsusp: It is, but the tests need to be tested with a framework. So I picked Keras.

test/functional/test_keras_framework.py
mvsusp: SageMaker containers is independent of any specific framework. A framework is just needed for testing. These tests don't run in any container as well.

setup.py
mvsusp: I agree with you. We can add other examples that are not machine learning related.
But I would ask why not using a machine learning specific example.

test/functional/test_keras_framework.py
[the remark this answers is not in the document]
yuanzhua: Thanks

test/functional/test_keras_framework.py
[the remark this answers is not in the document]
yuanzhua: I see

setup.py
nadiaya: Additional dependency that we don't really need, can be slower than just a fake machine learning script, implies necessity to know keras to anyone who wants to edit this package (it's going to be open sourced after all)

mvsusp: I like the idea to use a real framework in the tests. And code snippets of different frameworks in the documentation.

I agree that I will require longer to install the dependencies, but the required understand of Keras to understand the tests is extremely minimal.

I proposed us to keep the way it is for now. I am happy to change it if evidences that is not a good approach appear or more engineer agree with @nadiaya suggest approach. Please, let me know if that work for you, @nadiaya

laurenyu: In addition to the points Nadia brought up, I'd also argue that including Keras distracts from the purpose of the test - the test is just for determining if the container code is interacting correctly with the user script. What if there's a bug introduced in Keras (especially since we're not freezing on a specific version of keras here) - should our tests be dependent on that?

mvsusp: Keras is frozen.

PR #40, opened 2018-05-02
Do dynamic path expansion at runtime.

iquintero:
Instead of evaluating the custom paths at module import time, expand
the paths when the functions that use them get called. This also helps
making testing simpler as it avoids doing module reloads which have
nasty side effects.

*Issue #, if available:*

*Description of changes:*


winstonaws:
Can we just remove the BASE_PATH_ENV logic altogether and add it only if we need it? I can't think of a good use case for it right now.

PR #41, opened 2018-05-02
Simplify tests

mvsusp:
- not reloading module anymore to avoid inheritance errors
- remove dependency from keras in the tests
- simplified tests

It is the same PR that @winstonaws and @iquintero approved before, plus the keras removal and fix for the loading modules issue.


src/sagemaker_containers/environment.py
```diff
@@ -37,6 +40,7 @@
 BASE_PATH_ENV = 'BASE_PATH'  # type: str
 CURRENT_HOST_ENV = 'CURRENT_HOST'  # type: str
 JOB_NAME_ENV = 'JOB_NAME'  # type: str
+USE_NGINX_ENV = 'SAGEMAKER_USE_NGINX'  # type: str
```
iquintero: ? this is already defined below. Seems like the merge was not clean.

src/sagemaker_containers/environment.py
```diff
@@ -51,6 +55,11 @@
 RESOURCE_CONFIG_FILE = 'resourceconfig.json'  # type: str
 INPUT_DATA_CONFIG_FILE = 'inputdataconfig.json'  # type: str
 
+HYPERPARAMETERS_PATH = os.path.join(INPUT_CONFIG_PATH, HYPERPARAMETERS_FILE)  # type: str
+INPUT_DATA_CONFIG_FILE_PATH = os.path.join(INPUT_CONFIG_PATH, INPUT_DATA_CONFIG_FILE)  # type: str
+RESOURCE_CONFIG_PATH = os.path.join(INPUT_CONFIG_PATH, RESOURCE_CONFIG_FILE)  # type: str
+
+PROGRAM_PARAM = 'sagemaker_program'  # type: str
```
iquintero: same as above, USER_PROGRAM_PARAM is defined just below this line.

src/sagemaker_containers/environment.py
```diff
@@ -679,3 +689,24 @@ def model_server_workers(self):  # type: () -> int
         """Returns:
             (int): Number of worker processes the model server will use"""
         return self._model_server_workers
+
+
+@contextlib.contextmanager
+def temporary_directory(suffix='', prefix='tmp', dir=None):  # type: (str, str, str) -> None
```
iquintero: should this temporary_directory live in environment? I dont think so, this seems to be a generic context manager for temp dirs and I can see that its used at least in modules.py as well. Maybe functions.py?

test/miniml.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
iquintero: minimal.py? or is miniml.py intended?

test/unit/test_environment.py
```diff
@@ -20,7 +20,7 @@
 import six
 
 import sagemaker_containers as smc
-from test.conftest import json_dump
+import test
```
iquintero: Can we not use test?

> 
> The test package is meant for internal use by Python only. It is documented for the benefit of the core developers of Python. Any use of this package outside of Python’s standard library is discouraged as code mentioned here can change or be removed without notice between releases of Python.
> https://docs.python.org/3.5/library/test.html

src/sagemaker_containers/environment.py
mvsusp: I will fix it.

src/sagemaker_containers/environment.py
mvsusp: I will fix it.

src/sagemaker_containers/environment.py
mvsusp: Yeah, I thought about it. My thought process was that temporary files are part of the environment. Functions namespace only include functions that deal with functions. Do you have another suggestion?

laurenyu: maybe `tmp_dir` or `tmpdir` for the method name instead? or `create_tmpdir` if you're worried about overusing 'tmpdir'

test/miniml.py
mvsusp: It is a mock name for a framework. Do you prefer mini_ml ? Or mocked_framework ?

iquintero: ah, mini ml lol that works

src/sagemaker_containers/environment.py
iquintero: You missed this one.

mvsusp: I missed because they don't have the same name. Sorry about that.

PR #44, opened 2018-05-03
Add worker

mvsusp:


src/sagemaker_containers/worker.py
```diff
@@ -0,0 +1,93 @@
...
+    """Creates and Flask application from a transformer.
+
+    Args:
+        transform_fn (function): responsible to make predictions against the model. Follows the signature:
+
+            * Returns:
+                `sagemaker_containers.worker.TransformSpec`: named tuple with prediction data.
+
+
+        initialize_fn (function, optional): this function is called when the Flask application starts.
```
andremoeller: Should this default to the function that loads the model?
I want to avoid this case: my model takes a long time to load, but the server starts, so ping returns false 200s (when it can't accept inference requests yet). I think the default behavior should be to delay default_healthcheck_fn from returning 200s until the model is loaded

src/sagemaker_containers/worker.py
```diff
@@ -0,0 +1,93 @@
...
+            It doest not have return type or arguments.
+
+        healthcheck_fn (function, optional): function that will be used for healthcheck calls when the containers
+            starts, if not specified, it will use ping as the default healthcheck call. Signature:
+
+            * Returns:
+                `flask.app.Response`: response object with new healthcheck response.
+
+        module_name (str): the module name which implements the worker. If not specified, ir will use
```
andremoeller: "ir" -> "it"

src/sagemaker_containers/worker.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def invocations_fn():
+        transform_spec = transform_fn()
+
+        return Response(response=transform_spec.prediction,
+                        status=smc.status_codes.OK,
+                        mimetype=transform_spec.accept)
+
+    app.add_url_rule(rule='/invocations', endpoint='invocations', view_func=invocations_fn, methods=["POST"])
+    app.add_url_rule(rule='/ping', endpoint='ping', view_func=healthcheck_fn or default_healthcheck_fn)
```
andremoeller: Above, you have `healthcheck_fn = healthcheck_fn or default_healthcheck_fn`, so this can just be `view_func=healthcheck_fn`

(Or you can remove `healthcheck_fn = healthcheck_fn or default_healthcheck_fn`)

src/sagemaker_containers/worker.py
```diff
@@ -0,0 +1,93 @@
...
+    More information on how health-check works can be found here:
+    https://docs.aws.amazon.com/sagemaker/latest/dg/your-algorithms-inference-code.html#your-algorithms-inference-algo-ping-requests
+
+    Returns:
+        (flask.Response): with status code 200
+    """
+    return Response(status=smc.status_codes.OK)
+
+
+def run(transform_fn, initialize_fn=None, healthcheck_fn=None, module_name=None):
```
andremoeller: What about just doing healthcheck_fn=default_healthcheck_fn? Then the default is documented in the function signature rather than in the code, and you can avoid all assignments to `healthcheck_fn or default_healthcheck_fn`

test/functional/test_server.py
```diff
@@ -20,31 +20,37 @@
 
 
 def test_server():
```
andremoeller: Could you rename this test to be more specific (to describe what it tests)? like "test_server_conforms_to_hosting_schema"

src/sagemaker_containers/worker.py
mvsusp: A large model can take longer to load than then that time limit defined by SageMaker hosting to receive a successful ping (health check).  This is an issue already in the roadmap.

src/sagemaker_containers/worker.py
mvsusp: Good catch.

test/functional/test_server.py
mvsusp: Renamed to test_server_with_a_simple_app

PR #47, opened 2018-05-04
Add train cli.

iquintero:
train is the main entry point for training containers. The
framework_module should be set in the TrainingEnvironment and that is
all that is required to get started.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/sagemaker_containers/env.py
+++ b/src/sagemaker_containers/env.py
@@ -220,6 +220,8 @@ class Env(mapping.MappingMixin):
             num_cpu (int): The number of cpus available in the current container.
             module_name (str): The name of the user provided module.
             module_dir (str): The full path location of the user provided module.
+            framework_module (str):  Name of the framework module and entry point. For example:
+                my_module:main
     """
 
     def __init__(self):
@@ -304,7 +306,8 @@ class Env(mapping.MappingMixin):
     @property
     def framework_module(self):  # type: () -> str
         """Returns:
-            (str): Name of the framework module to be used for serving."""
+            (str): Name of the framework module and entry point. For example:
+                my_module:main"""
         return self._framework_module
 
     @staticmethod
@@ -630,7 +633,6 @@ class ServingEnv(Env):
             use_nginx (bool): Whether to use nginx as a reverse proxy.
             model_server_timeout (int): Timeout in seconds for the model server.
             model_server_workers (int): Number of worker processes the model server will use.
-            flask_app (str): Name of the flask app to use. Default: server:app
     """
 
     def __init__(self):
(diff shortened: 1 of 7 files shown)


src/sagemaker_containers/env.py
```diff
@@ -228,6 +228,7 @@ def __init__(self):
         module_dir = os.environ.get(SUBMIT_DIR_ENV, None)
         enable_metrics = util.strtobool(os.environ.get(ENABLE_METRICS_ENV, 'false')) == 1
         log_level = os.environ.get(LOG_LEVEL_ENV, logging.INFO)
+        framework_module = os.environ.get(FRAMEWORK_MODULE_ENV, ' ')
```
mvsusp: Should we fail fast here? What happens if the env vars is not passed? Or should we add a default value.

src/sagemaker_containers/env.py
```diff
@@ -299,6 +301,12 @@ def log_level(self):  # type: () -> int
         """
         return self._log_level
 
+    @property
+    def framework_module(self):  # type: () -> str
+        """Returns:
+            (str): Name of the framework module to be used for serving."""
```
mvsusp: What about: name of the framework module registered in pip or something like that?

src/sagemaker_containers/env.py
mvsusp: Nvm, I just noticed your todo

src/sagemaker_containers/trainer.py
```diff
@@ -0,0 +1,32 @@
...
+from sagemaker_containers import env, functions
+
+
+def train():
+    training_env = env.TrainingEnv()
+
+    # TODO: iquintero - add error handling for ImportError to let the user know
+    # if the framework module is not defined.
+    framework = importlib.import_module(training_env.framework_module)
+    model = framework.train(**functions.matching_args(framework.train, training_env))
```
mvsusp: The trainer should not be calling any framework functions. The trainer function is merely to invoke python -m framework_module:function_name so the framework entry point function for training is called.

The framework itself implement train and save functions. The reason of that is to provide freedom for the framework implementer to do use or not our functions.

test/unit/test_trainer.py
```diff
@@ -0,0 +1,41 @@
...
+# distributed on an 'AS IS' BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from mock import Mock, patch, PropertyMock
+
+from sagemaker_containers import env, trainer
+
+
+HYPERPARAMS = {
+    'x': 1,
```
mvsusp: What is this hyperparameter x for?

test/unit/test_trainer.py
```diff
@@ -0,0 +1,41 @@
...
+    'sagemaker_job_name': 'sagemaker-training-job',
+    'sagemaker_program': 'main.py',
+    'sagemaker_submit_directory': 'imagenet',
+    'sagemaker_enable_cloudwatch_metrics': True
+}
+
+
+@patch('importlib.import_module')
+@patch('sagemaker_containers.env.read_hyperparameters', lambda: HYPERPARAMS)
+@patch('sagemaker_containers.env.read_input_data_config', lambda: {'input': 'yes'})
```
mvsusp: Let's try to use real case values for the unit tests.

mvsusp: same here

PR #48, opened 2018-05-05
Add transformer

mvsusp:


src/sagemaker_containers/env.py
```diff
@@ -666,6 +667,64 @@ def model_server_workers(self):  # type: () -> int
         return self._model_server_workers
 
 
+class RequestEnv(Request, mapping.MappingMixin):
```
jesterhazy: this clearly models an HTTP request, so let's just call it Request.

src/sagemaker_containers/serializers.py
```diff
@@ -0,0 +1,185 @@
...
+import json as stdjson
+import textwrap
+
+import numpy as np
+from six import BytesIO, StringIO
+
+from sagemaker_containers import content_types
+
+
+class Npy(object):
```
jesterhazy: Why not NpySerializer (like the doc comment)?

src/sagemaker_containers/serializers.py
```diff
@@ -0,0 +1,185 @@
...
+
+        Returns:
+            (obj): serialized np array.
+        """
+        buffer = BytesIO()
+        np.save(buffer, nparray)
+        return buffer.getvalue()
+
+
+class Json(object):
```
jesterhazy: JsonSerializer. Soo many things called json or Json already.

src/sagemaker_containers/serializers.py
```diff
@@ -0,0 +1,185 @@
...
+        Args:
+            obj (object): object to be serialized.
+
+        Returns:
+            str:  serialized object
+        """
+        return self._numpy_dumps(obj) or self._stream_dumps(obj) or stdjson.dumps(obj)
+
+
+class Csv(object):
```
jesterhazy: ... Serializer

src/sagemaker_containers/serializers.py
```diff
@@ -0,0 +1,185 @@
...
+    """JSON serializer"""
```
jesterhazy: CSV

src/sagemaker_containers/serializers.py
```diff
@@ -0,0 +1,185 @@
...
+        obj (object): to be deserialized.
+        content_type (str): content type to be used.
+
+    Returns:
+        object: deserialized object.
+    """
+    return _SerializerMap(content_type).loads(obj)
+
+
+npy = Npy()
```
jesterhazy: module level... why?

src/sagemaker_containers/serializers.py
```diff
@@ -0,0 +1,185 @@
...
+
+        Returns:
+            str:  deserialized object.
+        """
+        stream = StringIO()
+        np.savetxt(stream, obj, delimiter=',', fmt='%s')
+        return stream.getvalue()
+
+
+def dumps(obj, content_type):
```
jesterhazy: top level... why? put them in a DefaultSerializer class.

src/sagemaker_containers/transformers.py
```diff
@@ -0,0 +1,160 @@
...
+
+    Examples:
+    >>>import os
+
+    >>>from sagemaker_containers import env, modules, transformers
+    >>>import miniml
+
+    >>>serving_env = env.ServingEnv()
+
+    >>>class MiniMlTransformer(transformers.BaseTransformer):
```
jesterhazy: MiniMl?

src/sagemaker_containers/transformers.py
```diff
@@ -0,0 +1,160 @@
...
+
+        Returns:
+
+        """
+        data = self.input_fn(input_data=input_data, content_type=content_type)
+        prediction = self.predict_fn(model=model, data=data)
+        return self.output_fn(prediction=prediction, accept=accept)
+
+    def initialize(self):
+        """function is called when the worker starts the Flask application.
```
jesterhazy: worker gets confusing here... i assume we are talking SM container worker class... but how does it related to gunicorn worker. especially important to be clear hear, since this function is all about startup order.

jesterhazy: also, they way i read this, it implies our default behavior is still to load model post-fork (once per worker). that's fine just asking.

src/sagemaker_containers/transformers.py
```diff
@@ -0,0 +1,160 @@
...
+        """
+        self._model = self.model_fn(model_dir=env.ServingEnv().model_dir)
+
+    def transform(self):
+        """Responsible to make predictions against the model.
+
+        Returns:
+            (sagemaker_containers.transformers.TransformSpec): named tuple with prediction data.
+        """
+        request_env = env.RequestEnv()
```
jesterhazy: again, this is really just Request not RequestEnv

src/sagemaker_containers/transformers.py
```diff
@@ -0,0 +1,160 @@
...
+
+        result = self.transform_fn(self._model, request_env.content, request_env.content_type, request_env.accept)
+
+        if not isinstance(result, TransformSpec):
+            # transforms tuple in TransformSpec for backwards compatibility
+            return TransformSpec(*result)
```
jesterhazy: TransformSpec seems pretty obscure. Why not just Result or Prediction? also, why not use a real class, instead of a named tuple, so we could hang some docs on it?

src/sagemaker_containers/worker.py
```diff
@@ -79,14 +77,11 @@ def run(transform_fn, initialize_fn=None, healthcheck_fn=None, module_name=None)
     def invocations_fn():
         transform_spec = transform_fn()
 
-        return Response(response=transform_spec.prediction,
+        return Response(response=transform_spec.serialized_prediction,
```
jesterhazy: the longer name doesn't seem to add much here

test/functional/test_transformer_implementation.py
```diff
@@ -0,0 +1,65 @@
...
+import os
+
+import numpy as np
+
+from sagemaker_containers import content_types, env, serializers, status_codes, transformers, worker
+import test
+from test import miniml
+
+
+class MiniMlTransformer(transformers.BaseTransformer):
```
jesterhazy: MiniMl?

test/miniml.py
```diff
@@ -27,3 +30,10 @@ def fit(self, x, y, **kwargs):
 
     def save(self, model_dir):
         test.write_json(self.parameters, model_dir)
+
+    @classmethod
+    def load(cls, model_dir):
+        return cls(**env.read_json(model_dir))
```
jesterhazy: why splat dict just to assign as dict (at 24)?

test/functional/test_transformer_implementation.py
```diff
@@ -0,0 +1,65 @@
...
+    def predict_fn(self, model, data):
+        return model.predict(data)
+
+    def model_fn(self, model_dir):
+        return miniml.Model.load(os.path.join(model_dir, 'minimlmodel'))
+
+
+def test_prediction_without_user_module():
```
jesterhazy: is there a test _with_ user module?

test/unit/test_serializers.py
```diff
@@ -0,0 +1,102 @@
...
+
+    string_io.return_value.getvalue.assert_called()
+    savetxt.assert_called_with(string_io(), 42, delimiter=',', fmt='%s')
+
+
+@patch('sagemaker_containers.serializers.npy.dumps', autospec=True)
+@patch('sagemaker_containers.serializers.csv.dumps', autospec=True)
+@patch('sagemaker_containers.serializers.json.dumps', autospec=True)
+def test_dumps(json_dumps, csv_dumps, npy_dumps):
+    serializers.dumps(42, content_types.JSON)
```
jesterhazy: just loop over content type/serializer tuples?

test/unit/test_serializers.py
```diff
@@ -0,0 +1,102 @@
...
+    serializers.dumps(42, content_types.NPY)
+
+    npy_dumps.assert_called_once_with(42)
+
+
+@patch('sagemaker_containers.serializers.npy.loads', autospec=True)
+@patch('sagemaker_containers.serializers.csv.loads', autospec=True)
+@patch('sagemaker_containers.serializers.json.loads', autospec=True)
+def test_loads(json_loads, csv_loads, npy_loads):
+    serializers.loads(42, content_types.JSON)
```
jesterhazy: loop?

test/functional/test_transformer_implementation.py
```diff
@@ -0,0 +1,65 @@
...
+
+        response = post(client, payload, content_types.NPY)
+
+        assert response.status_code == status_codes.OK
+        response_data = serializers.loads(response.get_data(), content_types.NPY)
+
+        np.testing.assert_array_almost_equal(response_data, np.asarray([3.6, 81., 1764.]))
+
+
+def post(client, payload, content_type):
```
jesterhazy: no accept param. do we allow mixed types (e.g. npy in/json out)? if not why not?

test/functional/test_transformer_implementation.py
```diff
@@ -0,0 +1,65 @@
...
+
+    transformer = MiniMlTransformer()
+    transformer.initialize()
+
+    with worker.run(transformer.transform, transformer.initialize, module_name='miniml').test_client() as client:
+        payload = [.6, 9, 42.]
+        response = post(client, payload, content_types.JSON)
+
+        assert response.status_code == status_codes.OK
+        assert response.get_data().decode('utf-8') == '[3.5999999999999996, 81.0, 1764.0]'
```
jesterhazy: potentially flaky test due to float precision?

src/sagemaker_containers/serializers.py
```diff
@@ -0,0 +1,185 @@
...
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the 'license' file accompanying this file. This file is
+# distributed on an 'AS IS' BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from __future__ import absolute_import
+
+import json as stdjson
```
jesterhazy: we should not rename this, and wouldn't have to if we didn't declare a module level json object of our own

test/functional/test_transformer_implementation.py
```diff
@@ -0,0 +1,65 @@
...
+
+        response = post(client, payload, content_types.CSV)
+
+        assert response.status_code == status_codes.OK
+        assert response.get_data().decode('utf-8') == '3.6000001430511475\n81.0\n1764.0\n'
```
jesterhazy: potentially flaky test due to float precision? also, weird that this is a different result than the json test above. obvi diff's between json and numpy internal conversion code.

test/functional/test_transformer_implementation.py
jesterhazy: but ok w/ deferring fix until we see a failure in the wild.

test/__init__.py
```diff
@@ -22,6 +22,7 @@
 import pytest
 import sagemaker
 import six
+import werkzeug.test as werkzeug_test
 
 # loading base path before loading the environment so all the environment paths are loaded properly
 os.environ['BASE_PATH'] = os.path.join(tempfile.mkdtemp(), 'opt', 'ml')
```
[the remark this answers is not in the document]
mvsusp: This is outside the scope of this PR, Jonathan. I am happy to talk with you and find ways to change that offline.

src/sagemaker_containers/env.py
```diff
@@ -666,6 +667,64 @@ def model_server_workers(self):  # type: () -> int
...
+    """The Request object used to read request data.
+
+    Example:
+
+    POST /INVOCATIONS {payload: '42', Content-Type: JSON, Accept: CSV}
```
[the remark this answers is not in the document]
mvsusp: Thanks!

src/sagemaker_containers/env.py
mvsusp: Yeah, I prefer Request as well. I will rename it.

src/sagemaker_containers/serializers.py
mvsusp: The namespace is being used to reduce the ambiguity:
```
from sagemaker_containers import serializers

serializers.Npy
```

```serializers.NpySerializer is just to big```

The previous idea would be: 
```
import sagemaker_containers as smc

smc.NpySerializer

```

Following PEP 8, PEP 432, and google style code. Ideally it should be:

```import sgmkr # or any small name that we come with

sgmk.serializer.Npy
```

@winstonaws is reaching out to the python community to understand better the best solution.

We scheduled a meeting specifically for this topic next Wednesday

src/sagemaker_containers/serializers.py
mvsusp: JsonSerializer is not Pythonic. The module (namespace) solves that problem for you.

from sagemaker_container.serializers import Json

Is not approved by PEP 8, Goggle style guide, and fails on linters.

src/sagemaker_containers/serializers.py
mvsusp: same

src/sagemaker_containers/serializers.py
mvsusp: Thanks

src/sagemaker_containers/serializers.py
mvsusp: Classes why? Look how python does: https://github.com/python/cpython/blob/master/Lib/json/__init__.py#L183

src/sagemaker_containers/serializers.py
mvsusp: Yeah, I am not sold on that one. Trying to think a cleaner solution.

src/sagemaker_containers/transformers.py
mvsusp: Just a fake example. I will change it to a Keras example

src/sagemaker_containers/transformers.py
mvsusp: Agreed.

src/sagemaker_containers/serializers.py
mvsusp: Can you point me to the PEP, lint, or code standard that says to not do that?

src/sagemaker_containers/transformers.py
mvsusp: @winstonaws suggested to return named tuples instead of tuples in the framework. @laurenyu and I agreed with the idea. Follow the conversation here: https://github.com/aws/sagemaker-containers/pull/36#discussion_r185141239

This is a best practice followed by stdlib and tensorflow. For example here https://github.com/python/cpython/blob/8a387219bdfb6ee34928d6168ac42ca559f11c9a/Lib/inspect.py#L1079

src/sagemaker_containers/worker.py
mvsusp: I agree.

test/functional/test_transformer_implementation.py
mvsusp: Removed

test/functional/test_transformer_implementation.py
mvsusp: This is a test helper function.

test/miniml.py
mvsusp: removed

mvsusp: Because kwargs is different than a dict:

from terminal
```
In [1]: def test(**kwargs):
   ...:     pass
   ...:
   ...: test({1:2})
   ...:
   ...:
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-1-e3da7728fca3> in <module>()
      2     pass
      3
----> 4 test({1:2})

TypeError: test() takes 0 positional arguments but 1 was given
```

test/__init__.py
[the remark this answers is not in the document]
jesterhazy: 👍

src/sagemaker_containers/serializers.py
jesterhazy: or `...serializer.NpySerializer`

src/sagemaker_containers/serializers.py
jesterhazy: I disagree that the namespace solves this, because after you import, you end up with a bunch of similarly named things.

how would ...serializers.JsonSerializer violate PEP8 or fail a lint check?

test/miniml.py
jesterhazy: i mean, change the __init__ method to take a dict, and just pass the dict here.

test/functional/test_transformer_implementation.py
jesterhazy: yes, if we support the mixed in/out

src/sagemaker_containers/serializers.py
jesterhazy: it's confusing... people reading are likely to know what the 'json' lib does, and won't know what stdjson is unless they go back and look at imports.

test/functional/test_transformer_implementation.py
jesterhazy: removed the test entirely? it's a good thing to test!

test/unit/test_serializers.py
mvsusp: Sure:

```
@pytest.mark.parametrize('target, content_type', [
    ('sagemaker_containers.serializers.json.dumps', content_types.JSON),
    ('sagemaker_containers.serializers.csv.dumps', content_types.CSV),
    ('sagemaker_containers.serializers.npy.dumps', content_types.NPY)
])
def test_dumps(target, content_type):
    with patch(target) as serializer:
        serializers.dumps(42, content_type)

        serializer.assert_called_once_with(42)


@pytest.mark.parametrize('target, content_type', [
    ('sagemaker_containers.serializers.json.loads', content_types.JSON),
    ('sagemaker_containers.serializers.csv.loads', content_types.CSV),
    ('sagemaker_containers.serializers.npy.loads', content_types.NPY)
])
def test_dumps(target, content_type):
    with patch(target) as serializer:
        serializers.loads(42, content_type)

        serializer.assert_called_once_with(42)
```

test/unit/test_serializers.py
mvsusp: yeah.

test/functional/test_transformer_implementation.py
mvsusp: removed

test/functional/test_transformer_implementation.py
mvsusp: There is, in the unit/test_transfomers. I forgot to rename this test.

test/functional/test_transformer_implementation.py
mvsusp: Is just a fake framework for the tests. @nadiaya and @laurenyu suggested me to not user Keras for the tests.

src/sagemaker_containers/worker.py
mvsusp: But prediction is miss leading as well. What do you suggest @jesterhazy ?

src/sagemaker_containers/transformers.py
mvsusp: I rephrased to:
```
    def initialize(self):  # type: () -> None
        """Execute any initialization necessary to starting making predictions with the Transformer.

        The default implementation is used to load the model.

        This function is called by sagemaker_containers.worker.run, before starting the Flask application.
        The gunicorn server forks multiple workers, executing multiple Flask applications in parallel.
        This function will be called once per each worker.

        It does not have return type or arguments.
        """
```

Should we rename this function to prepare? Or something else?

src/sagemaker_containers/serializers.py
mvsusp: ```from sagemaker_container.serializers import Json``` is not approved:

https://google.github.io/styleguide/pyguide.html#Imports

Use imports for packages and modules only.

src/sagemaker_containers/serializers.py
mvsusp: Looking at the imports is just a normal thing to do. Do you have any suggestion besides the unnecessary big statement  serializer.JsonSerializer().dump ? Remember that because of things like that, nobody likes Java.

mvsusp:
@jesterhazy what about the following:

sagemaker_containers.marshal.dumps

sagemaker_containers.marshal.json_dumps

sagemaker_containers.marshal.npy_loads 

?

src/sagemaker_containers/serializers.py
jesterhazy: this is unrelated to the package/class name issue. the google style guide specifically says not to use `import ... as` to rename a package unless it's "inconveniently long" which `json` definitely isn't. the only reason we "need" the rename is to avoid collision with the module level variable name json, which the google style also discourages.

src/sagemaker_containers/serializers.py
jesterhazy: yes, it should be:

```
import sagemaker_container.serializers 
...
# at point of use
serializer = sagemaker_container.serializers.JsonSerializer()
```
if a user (not us) finds that "inconveniently long", they can add an alias

```
import sagemaker_container.serializers as ser # (or whatever)

serializer = ser.JsonSerializer()
```

jesterhazy:
"marshal" is not meaningfully shorter than "serializer", and it's less clear.

src/sagemaker_containers/serializers.py
jesterhazy: eliminate the declarations. create the instances inside _SerializerMap.__init__ (but see comment on that class below)

src/sagemaker_containers/serializers.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+class _SerializerMap(object):
+    map = {content_types.JSON: json,
+           content_types.CSV: csv,
+           content_types.NPY: npy}
+
+    def __init__(self, content_type):
+        try:
+            self._serializer = self.map[content_type]
```
jesterhazy: this class doesn't need to be stateful. in fact i think we don't need it at all. instead create a DefaultSerializer class, move the module level serialization functions into it, and make this mapping function a (stateless) private function inside that class.

src/sagemaker_containers/encoder.py
```diff
@@ -0,0 +1,175 @@
...
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the 'license' file accompanying this file. This file is
+# distributed on an 'AS IS' BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from __future__ import absolute_import
+
+import json
```
mvsusp: @jesterhazy using normal json

src/sagemaker_containers/encoder.py
```diff
@@ -0,0 +1,175 @@
...
+import textwrap
+
+import numpy as np
+from six import BytesIO, StringIO
+
+from sagemaker_containers import content_types
+
+
+class NpyEncoder(object):
```
mvsusp: Using classes for encoder and decoders.

mvsusp:
@jesterhazy, I made the suggested changes. Unfortunately, the PR got even bigger.

To alleviate the review, I added comments where your previous comments were handled.
Thanks.

src/sagemaker_containers/transformers.py
```diff
@@ -0,0 +1,167 @@
...
+        """
+        request = worker.Request()
+
+        result = self.transform_fn(self._model, request.content, request.content_type, request.accept)
+
+        if not isinstance(result, worker.Response):
+            # transforms tuple in Response for backwards compatibility
+            return worker.Response(response=result[0], accept=result[1])
+
+        return result
```
mvsusp: Now that the Worker is a Flask application, this transform is basically a view functions: It creates the request object, makes the prediction and returns a Response object.

src/sagemaker_containers/encoder.py
```diff
@@ -0,0 +1,175 @@
...
+            obj (object): object to be deserialized.
+
+        Returns:
+            (np.array): deserialized numpy array.
+        """
+        stream = BytesIO(obj)
+        return np.load(stream)
+
+
+class JsonEncoder(json.JSONEncoder):
```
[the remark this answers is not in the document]
jesterhazy: 👍

src/sagemaker_containers/encoder.py
```diff
@@ -0,0 +1,175 @@
...
+        """Encode an object to a one of the default content types.
+
+        Args:
+            obj (object): to be encoded.
+            content_type (str): content type to be used.
+
+        Returns:
+            object: encoded object.
+        """
+        for encoder in self._encoders:
```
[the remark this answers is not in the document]
mvsusp: The number of encoders are 3, not sure if one day will be 10, which is why I did not use the map. I happy to add if you want me to, but I don't see perf gain here.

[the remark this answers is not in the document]
jesterhazy: absolutely need that map... otherwise we iterate over this list, do a function call and a compare on each item *every time a request is handled*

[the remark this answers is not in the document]
mvsusp:
```
for encoder in self._encoders:
    if content_type == encoder.content_type:
        return encoder.encode(obj)

```

for the case that _encoders are 3 is the same of:

```
if content_type == JSON:
    return encoder.encode(obj)
if content_type == CSV:
    return encoder.encode(obj)
if content_type == NPY:
    return encoder.encode(obj)
```

if we don't consider  any optimizations that the interpreter does with if statements.

The encode function call only happens once (Not different than the map). If you want to optimize this path even more, the 3 chained if elses will be slighter faster than the map.

I will change to dict, but again, I am not sure how substantial is the improvement (I believe is very little).

src/sagemaker_containers/encoder.py
```diff
@@ -0,0 +1,175 @@
...
+
+        Returns:
+            str:  decoded object.
+        """
+        stream = StringIO()
+        np.savetxt(stream, obj, delimiter=',', fmt='%s')
+        return stream.getvalue()
+
+
+DEFAULT_ENCODERS = [JsonEncoder(), CsvEncoder(), NpyEncoder()]
```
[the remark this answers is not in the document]
jesterhazy: ok, it can be a list, but then DefaultDecoder should turn it into a dict in its constructor. also, allowing people to pass in a list doesn't mean we need to expose our list as a (mutable!) module variable. it could just be default in DefaultDecoder's __init__ if no list is supplied.

src/sagemaker_containers/transformers.py
laurenyu: (to be fair, I'm not opposed to a real class instead of a named tuple. just wasn't a fan of the unnamed tuple)

src/sagemaker_containers/encoders.py
```diff
@@ -138,15 +140,20 @@ def encode(self, obj, content_type):
...
-            if content_type == encoder.content_type:
-                return encoder.encode(obj)
-        raise UnsupportedFormatError(content_type)
+        try:
+            return self._encoders_map[content_type].encode(obj)
+        except KeyError:
+            raise UnsupportedFormatError(content_type)
+
+
+default_encoder = DefaultEncoder()
```
jesterhazy: remove module var

src/sagemaker_containers/encoders.py
```diff
@@ -158,10 +165,13 @@ def decode(self, obj, content_type):
...
-            if content_type == decoder.content_type:
-                return decoder.decode(obj)
-        raise UnsupportedFormatError(content_type)
+        try:
+            return self._decoders_map[content_type].decode(obj)
+        except KeyError:
+            raise UnsupportedFormatError(content_type)
+
+
+default_decoder = DefaultDecoder()
```
jesterhazy: remove module var

src/sagemaker_containers/transformers.py
```diff
@@ -85,7 +90,7 @@ def input_fn(input_data, content_type):
         Returns:
             (obj): data ready for prediction.
         """
-        return encoder.DefaultDecoder().decode(input_data, content_type)
+        return encoders.default_decoder.decode(input_data, content_type)
```
jesterhazy: make self.decoder attr in this class's __init__ method instead of using module var in encoders

src/sagemaker_containers/transformers.py
```diff
@@ -108,14 +113,13 @@ def output_fn(prediction, accept):
...
-            (TransformSpec): a namedtuple with the following args:
+            (worker.Response): a Flask response object with the following args:
 
                 * Args:
                     response: the serialized data to return
                     accept: the content-type that the data was transformed to.
         """
-        default_encoder = encoder.DefaultEncoder()
-        return worker.Response(default_encoder.encode(prediction, accept), accept)
+        return worker.Response(encoders.default_encoder.encode(prediction, accept), accept)
```
jesterhazy: make self.encode attr in __init__ instead

test/__init__.py
```diff
@@ -54,7 +54,7 @@ def write_json(obj, path):  # type: (object, str) -> None
         path (str): Path to JSON file
     """
     with open(path, 'w') as f:
-        json.dump(obj, f)
+        json.dump(obj, f, )
```
jesterhazy: ?

src/sagemaker_containers/transformers.py
jesterhazy: discussed offline. ok as is.

src/sagemaker_containers/encoders.py
jesterhazy: discussed offline. ok as is.

src/sagemaker_containers/encoders.py
jesterhazy: discussed offline. ok as is.

src/sagemaker_containers/transformers.py
jesterhazy: discussed offline. ok as is.

test/__init__.py
jesterhazy: @mvsusp i think this still needs attention

src/sagemaker_containers/encoders.py
mvsusp: @jesterhazy found the following in the google style guide:

[05/07/18 11:47 AM] Esterhazy, Jonathan: Global variables
link ▽ Avoid global variables.
Definition:
Variables that are declared at the module level.

Pros:
Occasionally useful.

Cons:
Has the potential to change module behavior during the import, because assignments to module-level variables are done when the module is imported.

Decision:
Avoid global variables in favor of class variables. Some exceptions are:

Default options for scripts.
Module-level constants. For example: PI = 3.14159. Constants should be named using all caps with underscores; see Naming below.
It is sometimes useful for globals to cache values needed or returned by functions.
If needed, globals should be made internal to the module and accessed through public module level functions; see Naming below.
[05/07/18 11:47 AM] Esterhazy, Jonathan: ttps://google.github.io/styleguide/pyguide.html?showone=Global_variables#Global_variables

mvsusp: I am working in a solution outside this PR. I will add a linter as well, to better rule enforcement.
@jesterhazy is plans to work in a best practices guide for the team.

PR #52, opened 2018-05-11
Support to create a Transformer from functions

mvsusp:
# Description of changes:

Support to create a Transformer from user functions.


src/sagemaker_containers/transformer.py
```diff
@@ -0,0 +1,162 @@
...
+    """Function responsible for model predictions.
+
+    Args:
+        model (obj): model loaded by model_fn
+        data: de-serializes data returned by input_fn
+
+     Returns:
+         (obj): data ready for prediction.
+    """
+    raise NotImplementedError()
```
winstonaws: Should include an error message in the exception.

src/sagemaker_containers/functions.py
```diff
@@ -66,3 +66,22 @@ def getargspec(fn):  # type: (function) -> inspect.ArgSpec
...
+    Returns:
+        (object): fn wrapped in a try catch.
+    """
+    def wrapper(*args, **kwargs):
+        try:
+            return fn(*args, **kwargs)
+        except Exception as e:
+            six.raise_from(error_class(e), e)
+
+    return fn and wrapper
```
winstonaws: I wouldn't have this branch to return None if None is passed in, since it still doesn't validate correctly if any other type is passed in. If you want you could validate if the passed in fn is callable instead. See note below in Transformer init as well, since this affects that.

src/sagemaker_containers/transformer.py
```diff
@@ -0,0 +1,162 @@
...
+
+        Args:
+            model_fn (fn): Function responsible to load the model.
+            input_fn (fn): Takes request data and de-serializes the data into an object for prediction.
+            predict_fn (fn): Function responsible for model predictions.
+            output_fn (fn): Function responsible to serialize the prediction for the response.
+            error_class (Exception): Error class used to separate framework and user errors.
+        """
+        self._model = None
+        self._model_fn = functions.error_wrapper(model_fn, error_class) or default_model_fn
```
winstonaws: This line is unintuitive since I wouldn't expect error_wrapper to return None if model_fn is None. I think it'd be bettter like:

functions.error_wrapper(model_fn, error_class) if model_fn else default_model_fn

src/sagemaker_containers/transformer.py
```diff
@@ -0,0 +1,162 @@
...
+    >>>def model_fn(model_dir):
+    >>>     return Keras.models.load_model(os.path.join(model_dir, 'minimlmodel'))
+    >>>
+    >>>transformer = transformer.Transformer(predict_fn=predict_fn, model_fn=model_fn)
+    >>>
+    >>>mod = modules.download_and_import(serving_env.module_dir, serving_env.module_name)
+    >>>transformer.load_user_fns(mod)
+    """
+
+    def __init__(self, model_fn=None, input_fn=None, predict_fn=None, output_fn=None, error_class=ClientError):
```
winstonaws: Are we not allowing transform_fn to be loaded from the user module? 

If we are, shouldn't it be a constructor argument?

src/sagemaker_containers/functions.py
mvsusp: Yeah... I was hoping we could find a more elegant alternative than this. I feel that we should not even check if it None or not here... any thoughts?

src/sagemaker_containers/transformer.py
mvsusp: Yeah... Maybe that is the solution here. I wanted to separate creating the FrameworkTransformer and extracting the user functions.

My intention is to optionally allow customers to do the following:

*Let's suppose the following user script*
```python
from sagemaker_chainer import worker 

def predict_fn(...):
    pass

if __name__ == '__main__':
    worker.from_functions(predict_fn=predict_fn).run()

src/sagemaker_containers/transformer.py
mvsusp: Yeah, I agree.

src/sagemaker_containers/functions.py
winstonaws: Yea, I think either we should do full validation (check if its callable), or no validation. We shouldn't special case None. 

If we change the init to use the slightly more verbose " ... if ... else ... " syntax then we don't need the None case.

mvsusp: Got it, in the comment below. Let's keep it simple.

PR #60, opened 2018-05-16
Support to requirements.txt file

mvsusp:
# Description of changes:

- Create a separate module for errors
- Install dependencies via requirements.txt or setup.py provide with the user script
-  Fix download and import tests
- Install additional data files to the package

### DIFF
--- a/src/sagemaker_containers/modules.py
+++ b/src/sagemaker_containers/modules.py
@@ -19,12 +19,13 @@ import shlex
 import subprocess
 import sys
 import tarfile
-import traceback
+import textwrap
 
 import boto3
+import six
 from six.moves.urllib.parse import urlparse
 
-from sagemaker_containers import env
+from sagemaker_containers import env, errors
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


src/sagemaker_containers/modules.py
```diff
@@ -77,9 +105,14 @@ def install(path):  # type: (str) -> None
     if not sys.executable:
         raise RuntimeError('Failed to retrieve the real path for the Python executable binary')
     try:
-        subprocess.check_call(shlex.split('%s -m pip install %s -U' % (sys.executable, path)))
-    except subprocess.CalledProcessError:
-        raise RuntimeError('Failed to pip install %s:%s%s' % (path, os.linesep, traceback.format_exc()))
+        cmd = '%s -m pip install -U . ' % sys.executable
```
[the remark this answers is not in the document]
laurenyu: the issue last time was that the logic for downloading the user code was in the TF container, and the call to install requirements occurred before that downloading happened.

test/functional/test_download_and_import.py
```diff
@@ -12,45 +12,106 @@
...
 
 
-def test_download_and_import_script_with_requirements():
+def test_download_and_import_script_with_requirements(user_module_name):
     user_module = test.UserModule(USER_SCRIPT_WITH_REQUIREMENTS).add_file(REQUIREMENTS_FILE).upload()
 
-    module = modules.download_and_import(user_module.url, 'test_script')
+    module = modules.download_and_import(user_module.url, user_module_name, cache=False)
+
+    # flake8: noqa ignore=W291 trailing whitespace
```
laurenyu: why do you need trailing whitespace here? does this mean things will break if someone's editor automatically deletes all trailing whitespace?

mvsusp: Good call. I will remove the noqa

PR #61, opened 2018-05-16
Backward compatibility support

mvsusp:
# Description of changes:

- add network_interface_name env parameter
- append current_host to output_data_dir
- revert parameters to predict_fn to match MXNet
- invoking transformation functions with positional arguments instead of named arguments
- unicode support
- passing dtype in the encoder functions

### DIFF
--- a/src/sagemaker_containers/encoders.py
+++ b/src/sagemaker_containers/encoders.py
@@ -51,6 +51,23 @@ def npy_to_numpy(npy_array):  # type: (object) -> np.array
     return np.load(stream)
 
 
+def unicode_to_str(string_like):  # type: (str or unicode) -> str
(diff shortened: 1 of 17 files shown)


src/sagemaker_containers/encoders.py
```diff
@@ -38,12 +38,13 @@ def array_to_npy(array_like):  # type: (np.array or Iterable or int or float) ->
     return buffer.getvalue()
 
 
-def npy_to_numpy(npy_array):  # type: (object) -> np.array
+def npy_to_numpy(npy_array, dtype=np.float32):  # type: (object) -> np.array
```
jesterhazy: remove new argument (not used in function body)

PR #62, opened 2018-05-17
Fix functional test

mvsusp:
## Description of changes:

- Remove cache from the test so module is re installed
- Remove module after training

### DIFF
--- a/test/functional/test_training_framework.py
+++ b/test/functional/test_training_framework.py
@@ -13,6 +13,8 @@
 from __future__ import absolute_import
 
 import os
+import shlex
+import subprocess
 
 import numpy as np
 import pytest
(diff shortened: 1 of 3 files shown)


test/functional/test_training_framework.py
```diff
@@ -75,17 +88,17 @@ def framework_training_fn():
             model.save(model_file)
 
 
-@pytest.mark.parametrize('user_script', [USER_SCRIPT, USER_SCRIPT_WITH_SAVE])
-def test_training_framework(user_script):
+def test_training_framework_with_save():
```
yangaws: Why do we give up using pytest.mark.parametrize? Since there is likely to have more tests using more scripts, for example, trainer tests, I definitely prefer using pytest.mark.parametrize for more scalable style.

Does using pytest.mark.parametrize break caching?

yangaws: Talked offline. Will change this later.

PR #76, opened 2018-05-24
Creating Logger BasicConfig after importing the framework

mvsusp:
## Description of changes:

- Creating Logger BasicConfig after importing the framework

### DIFF
--- a/src/sagemaker_containers/_modules.py
+++ b/src/sagemaker_containers/_modules.py
@@ -24,7 +24,9 @@ import textwrap
 import boto3
 import six
 from six.moves.urllib.parse import urlparse
+import yaml
 
+import sagemaker_containers
 from sagemaker_containers import _errors, _files
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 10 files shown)


src/sagemaker_containers/_modules.py
```diff
@@ -24,7 +24,9 @@
 import boto3
 import six
 from six.moves.urllib.parse import urlparse
+import yaml
```
jesterhazy: we do json for other config data. why introduce yaml?

jesterhazy: oh i see, we are using it for formatting output. I think we should just stick with native dict repr.

mvsusp: Agreed, removed PYAML.

src/sagemaker_containers/beta/framework/__init__.py
```diff
@@ -19,6 +19,7 @@
 from sagemaker_containers import _errors as errors
 from sagemaker_containers import _env as env
 from sagemaker_containers import _functions as functions
+from sagemaker_containers import _logging as logging
 from sagemaker_containers import _mapping as mapping
 from sagemaker_containers import _modules as modules
```
jesterhazy: change is out of scope for this PR, but its weird that we renamed all these modules _ and then alias them back to the old public name.

PR #88, opened 2018-06-08
Make sure output_data_dir exists.

nadiaya:
Make sure output_data_dir exists.
Should fairly safe to call directory creation since we do not expect anyone else creating the directory at the same time as code is running.

No  tests added since I am not sure if can properly test it without running on EASE.


nadiaya:
Closing this pull request. Will post a different one based on discussed approach in a moment.

PR #94, opened 2018-07-03
Correct links to Tensorflow and MXNet containers

harusametime:
*Description of changes:*

Just correcting the links.


nadiaya:
Thank you so much for finding and addressing this error!

The Readme is actually correct - we were supposed to update repos names but somehow haven't done this.
I've updated the names of the repos and now old links should be working:
https://github.com/aws/sagemaker-tensorflow-container
https://github.com/aws/sagemaker-mxnet-container

PR #95, opened 2018-07-24
Support for local modules for SageMaker frameworks

ChoiByungWook:
*Description of changes:*
Added support for running user modules that are copied into the container / image for a predefined SageMaker framework container (pytorch, chainer).

Will add unit/integ tests after.

Will need to modify pytorch and chainer containers.


src/sagemaker_containers/_modules.py
```diff
@@ -152,19 +152,23 @@ def download_and_install(url, name=DEFAULT_MODULE_NAME, cache=True):
 
     if not should_use_cache:
         with _files.tmpdir() as tmpdir:
-            dst = os.path.join(tmpdir, 'tar_file')
-            s3_download(url, dst)
+            module_path = url
```
mvsusp: remove this variable.

src/sagemaker_containers/_env.py
```diff
@@ -222,7 +222,7 @@ class _Env(_mapping.MappingMixin):
     def __init__(self):
         current_host = os.environ.get(_params.CURRENT_HOST_ENV)
         module_name = os.environ.get(_params.USER_PROGRAM_ENV, None)
-        module_dir = os.environ.get(_params.SUBMIT_DIR_ENV, None)
+        module_dir = os.environ.get(_params.SUBMIT_DIR_ENV, 'file:///opt/ml/code')
```
mvsusp: Avoid using slash as directory separator because it does not work in all file systems. Use os.path.join

src/sagemaker_containers/_env.py
```diff
@@ -448,8 +448,9 @@ def __init__(self, resource_config=None, input_data_config=None, hyperparameters
         self._current_host = current_host
 
         # override base class attributes
-        self._module_name = str(sagemaker_hyperparameters[_params.USER_PROGRAM_PARAM])
-        self._module_dir = str(sagemaker_hyperparameters[_params.SUBMIT_DIR_PARAM])
+        if self._module_name is None:
+            self._module_name = str(sagemaker_hyperparameters[_params.USER_PROGRAM_PARAM])
+        self._module_dir = str(sagemaker_hyperparameters.get(_params.SUBMIT_DIR_PARAM, 'file:///opt/ml/code'))
```
mvsusp: It seems that opt/ml/code can be a constant. Consider that option.

src/sagemaker_containers/_modules.py
```diff
@@ -131,7 +131,7 @@ def exists(name):  # type: (str) -> bool
         return True
 
 
-def download_and_install(url, name=DEFAULT_MODULE_NAME, cache=True):
+def download_and_install(uri, name=DEFAULT_MODULE_NAME, cache=True):
     # type: (str, str, bool) -> module
     """Download, prepare and install a compressed tar file from S3 as a module.
```
mvsusp: Update Docstrings

src/sagemaker_containers/_env.py
mvsusp: Do we need to have **file** in the front? My reason is that it is requires to have remove the prefix file before opening the content.

src/sagemaker_containers/_modules.py
```diff
@@ -152,19 +152,25 @@ def download_and_install(url, name=DEFAULT_MODULE_NAME, cache=True):
...
+                module_path = os.path.join(tmpdir, 'module_dir')
+                os.makedirs(module_path)
 
-            module_path = os.path.join(tmpdir, 'module_dir')
+                with tarfile.open(name=dst, mode='r:gz') as t:
+                    t.extractall(path=module_path)
 
-            os.makedirs(module_path)
+            elif uri.startswith('file://'):
+                module_path = uri.replace('file://', '', 1)
```
mvsusp: See comment above about file://

src/sagemaker_containers/_env.py
ChoiByungWook: I was following the convention set forward here, https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/job.py#L104, for local mode support within the python sdk.

ChoiByungWook: The code that invokes this URI does drop the prefix. Good point on the directory separator.

src/sagemaker_containers/_modules.py
ChoiByungWook: Updated.

PR #96, opened 2018-07-26
Support to use SageMaker Containers outside SageMaker training

mvsusp:
*Description of changes:*
- SageMaker containers create a folder under ~/sagemaker_local/jobs with the required directories and files when used outside SageMaker Training 

### DIFF
--- a/src/sagemaker_containers/__init__.py
+++ b/src/sagemaker_containers/__init__.py
@@ -12,10 +12,6 @@
 # language governing permissions and limitations under the License.
 from __future__ import absolute_import
 
-from sagemaker_containers import _logging
-
-logger = _logging.get_logger()
-
 
 def training_env():
     """Create a TrainingEnv.
--- a/src/sagemaker_containers/__init__.py
+++ b/src/sagemaker_containers/__init__.py
@@ -12,6 +12,10 @@
 # language governing permissions and limitations under the License.
 from __future__ import absolute_import
 
+from sagemaker_containers import _logging
+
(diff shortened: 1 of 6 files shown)


src/sagemaker_containers/__init__.py
```diff
@@ -12,15 +12,21 @@
 # language governing permissions and limitations under the License.
 from __future__ import absolute_import
 
+from sagemaker_containers import _logging
+
+logger = _logging.get_logger()
```
ChoiByungWook: ?

src/sagemaker_containers/_env.py
```diff
@@ -84,6 +137,31 @@
...
+    logger.info('Creating a new training folder under %s .' % base_dir)
+
+    os.makedirs(model_dir)
+    os.makedirs(input_config_dir)
+    os.makedirs(output_data_dir)
+
+    _write_json({}, hyperparameters_file_dir)
+    _write_json({}, input_data_config_file_dir)
+
+    host_name = socket.gethostname()
```
ChoiByungWook: Will this work for every docker container?

mvsusp: I believe so. It works for Windows, Mac, and Unix https://docs.python.org/2/library/socket.html

src/sagemaker_containers/__init__.py
mvsusp: Good catch, I will remove it.

PR #105, opened 2018-08-22
Fix error message retrieval

laurenyu:
*Issue #, if available:*
https://github.com/aws/sagemaker-python-sdk/issues/362

*Description of changes:*
`e.message` isn't valid after something like Python 2.6, so replacing that with `str(e)`


nadiaya:
Make sure you run tests in python 2 and python 3

PR #107, opened 2018-09-21
Update boto client with region_name

yangaws:
*Issue #, if available:*

*Description of changes:*
Update boto client with region_name, use env variable 'AWS_REGION'

### DIFF
--- a/src/container_support/utils.py
+++ b/src/container_support/utils.py
@@ -11,6 +11,7 @@
 #  express or implied. See the License for the specific language governing 
 #  permissions and limitations under the License.
 
+import os
 import tarfile
 
 import boto3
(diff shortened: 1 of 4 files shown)


src/container_support/utils.py
```diff
@@ -30,7 +31,7 @@ def download_s3_resource(source, target):
     """ Downloads the s3 object source and stores in a new file with path target.
     """
     print("Downloading {} to {}".format(source, target))
-    s3 = boto3.resource('s3')
+    s3 = boto3.resource('s3', region_name=os.environ['AWS_REGION'])
```
ChoiByungWook: I think a os.environ.get() would be better. Since, this will fail if the container doesn't have this environment variable defined.

PR #122, opened 2018-10-23
Support user-supplied transform_fn for Transformer

laurenyu:
*Description of changes:*
Allows the `Transformer` object to be initialized with a custom `transform_fn` if none of the other three methods are defined. This user-supplied method must have the signature:

```python
def transform_fn(model, data, content_type, accept_type):
    pass
```

Its return value should be either a `worker.Response` object or a tuple that's `(data, accept_type)`

In addition to running tests here, I also made an MXNet container based on this change and ran a couple inference tests.

### DIFF
--- a/src/sagemaker_containers/_transformer.py
+++ b/src/sagemaker_containers/_transformer.py
@@ -135,8 +135,10 @@ class Transformer(object):
         if transform_fn and (input_fn or predict_fn or output_fn):
             raise ValueError('Cannot use transform_fn implementation with input_fn, predict_fn, and/or output_fn')
 
-        self._transform_fn = _functions.error_wrapper(transform_fn, error_class) \
-            if transform_fn else self._default_transform_fn
+        if transform_fn is not None:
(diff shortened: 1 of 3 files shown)


src/sagemaker_containers/_transformer.py
```diff
@@ -123,10 +124,20 @@ def __init__(self, model_fn=None, input_fn=None, predict_fn=None, output_fn=None
...
             error_class (Exception): Error class used to separate framework and user errors.
         """
         self._model = None
         self._model_fn = _functions.error_wrapper(model_fn, error_class) if model_fn else default_model_fn
+
+        if transform_fn and (input_fn or predict_fn or output_fn):
+            raise ValueError('Cannot use transform_fn implementation with input_fn, predict_fn, and/or output_fn')
+
+        self._transform_fn = _functions.error_wrapper(transform_fn, error_class) \
+            if transform_fn else self._default_transform_fn
```
mvsusp: nit: ternaries should be used only for one liners https://github.com/google/styleguide/blob/gh-pages/pyguide.md#2114-decision

PR #123, opened 2018-10-23
Make it optional for run_module to wait for subprocess to finish

icywang86rui:
*Issue #, if available:*

*Description of changes:*
This change also adds sagemaker_parameter_server_num to TrainingEnv. It will be
 used in the TensorFlow containers to launch parameter server process for users.
 The change to run_module is necessary because the parameter server process never
 exits so the main process can't wait for it to finish.


CHANGELOG.rst
```diff
@@ -3,9 +3,11 @@ CHANGELOG
 =========
 
 2.2.5dev
-=====
+========
```
laurenyu: just don't forget to remove the extra === when it comes time to release :) (that's why I didn't bother adding it)

src/sagemaker_containers/_env.py
```diff
@@ -550,6 +551,15 @@ def job_name(self):  # type: () -> str
         """
         return self._job_name
 
+    @property
+    def parameter_server_num(self):  # type: () -> int
```
laurenyu: is this useful for more than just TF?

icywang86rui: I think so. It's a possible use case for mxnet or pytorch, right? Chainer uses mpi no parameter servers there.

CHANGELOG.rst
icywang86rui: Will do.

src/sagemaker_containers/_env.py
laurenyu: MXNet is going to stay out of the user's way as much as possible, so I don't think this would be useful for the current MXNet plan.

For PyTorch, I don't see anything in the [training file](https://github.com/aws/sagemaker-pytorch-container/blob/master/src/sagemaker_pytorch_container/training.py) that would suggest being able to use this. It also does very little in terms of setup.

icywang86rui: sagemaker-containers filters out all the 'sagemaker_' prefixed hyperparameters. If I do this in sagemaker-tensorflow-containers I have to use a regular hyperparameter and remove it from the dict before calling the user_module. It's really hacky.

laurenyu: fair enough

src/sagemaker_containers/_env.py
```diff
@@ -522,6 +522,7 @@ def __init__(self, resource_config=None, input_data_config=None, hyperparameters
 
         self._hosts = hosts
         self._network_interface_name = network_interface_name
+        self._parameter_server_num = sagemaker_hyperparameters.get(_params.PARAMETER_SERVER_NUM_PARAM)
```
mvsusp: It's ok to do that in SageMaker containers if parameter servers is a common concept across all containers. We might have to change PyTorch and Chainer to be able to do that. I guess, we will have to get feedback from the team first.

src/sagemaker_containers/_env.py
mvsusp: @icywang86rui You can use the constructor to do the following (https://github.com/aws/sagemaker-containers/blob/c31296cc1b0f53dc0879f6d45d328c3d04d004e2/src/sagemaker_containers/beta/framework/__init__.py#L32):

```python

hps = env.read_hyperparameters()
env = create_training_env(hyperparameters=hps)

my_field = hps['field']
```
What can we do to improve this experience in a generic way?

src/sagemaker_containers/_modules.py
```diff
@@ -171,7 +171,7 @@ def download_and_install(uri, name=DEFAULT_MODULE_NAME, cache=True):
             install(module_path)
 
 
-def run(module_name, args=None, env_vars=None):  # type: (str, list, dict) -> None
+def run(module_name, args=None, env_vars=None, wait=True):  # type: (str, list, dict) -> Popen
```
mvsusp: Please update docstrings.

src/sagemaker_containers/_modules.py
```diff
@@ -285,7 +293,9 @@ def run_module(uri, args, env_vars=None, name=DEFAULT_MODULE_NAME, cache=True):
         args (list):  A list of program arguments.
         env_vars (dict): A map containing the environment variables to be written.
         name (str): name of the script or module.
-        cache (bool): if True it will avoid downloading the module again, if already installed.
+        cache (bool): If True it will avoid downloading the module again, if already installed.
+        wait (bool): If True run_module will wait for the user module to exit and check the exit code,
+                     otherwise it will launch the user module with subprocess and return the process object.
```
mvsusp: Nice

src/sagemaker_containers/_env.py
jesterhazy: can we change this so that we just capture all sagemaker_ parameters in a dict, store the dict in TrainingEnv in addition to the other named attributes? That lets us flexibly pass framework-specific stuff through this package without adding any framework-specific code here.

all sagemaker_ parameters are all captured in a dict and the

mvsusp: This is an option @jesterhazy . Add the time, we decided to not do that because the TrainingEnv was designed to be used by customer as well: providing env.sagemaker_hyperparameters could be confusing. We can review the decision.

An alternative of this option if subclassing: class TFEnv(TrainingEnv) in the sagemaker_tensorflow library with the helper functions.

icywang86rui: My original thought was this can be used as in other framework as well. I don't want to use the normal hyperparameter because that's for customer to use in the training script. If we are only going to do this for Tensorflow, it might make sense to use a different prefix and capture those and store in TrainingEnv like @jesterhazy mentioned.

I will make the change after our discussion this afternoon.

mvsusp: I like the idea to add this to all frameworks. Are you planning on making the changes in PyTorch and Chainer?

src/sagemaker_containers/_env.py
```diff
@@ -550,6 +555,16 @@ def job_name(self):  # type: () -> str
         """
         return self._job_name
 
+    @property
+    def additional_framework_parameters(self):  # type: () -> dict
+        """The dict of additional framework hyperparameters. All the hyperparameters prefixed with 'sagemaker_' but
+            not in SAGEMAKER_HYPERPARAMETERS will be included here.
+
+        Returns:
```
mvsusp:
```suggestion
                SageMaker Python SDK adds hyperparameters with a prefix **sagemaker_** during training. These hyperparameters are framework independent settings and are not defined by the user. 
                
        Returns:
```

PR #127, opened 2018-11-01
Stream stderr instead of capturing it when running training script

icywang86rui:
*Issue #, if available:*

*Description of changes:*
TensorFlow logs to stderr when training is being executed outside of notebook. Currently users won't see any output from TensorFlow in their training log. This change stream out stderr instead of capturing it. Also subprocess.communicate() caches the bytes in memory with long running training jobs the buffer is in danger of overflowing. This change mitigate this risk as well.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,11 @@
 CHANGELOG
 =========
 
+2.2.7
+=====
+* feature: Making pip install less noisy
+* bug-fix: Stream stderr instead of capturing it when running user script
+
 2.2.6
 =====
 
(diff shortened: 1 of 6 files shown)


icywang86rui:
Whatever got output to stderr will still be in the training log. This change fixes the case where the return_code is 0 and all the output stderr got swallowed. Ideally if the return_code is not 0 we might want to still stick the stderr content in the error message. But we won't know to capture or stream when the call is still executing. We can do both but for our use case the call will likely be long running so there is still the risk of overflowing the buffer.

CHANGELOG.rst
```diff
@@ -2,6 +2,11 @@
 CHANGELOG
 =========
 
+2.2.7
+=====
+* feature: Making pip install less noisy
```
laurenyu: add a newline between ll. 6-7

icywang86rui:
I will add the empty line when I post the pr to bump the version.

PR #130, opened 2018-11-08
Changes for bash

mvsusp:
These are main changes to allow customers to run bash scripts:
In addition to that, we need to:
- consider creating a parameter in the Python SDK Estimator or allow customers to pass a hyperparameter to the Python SDK Estimator
- create notebook examples, tests, etc

### DIFF
--- a/src/sagemaker_containers/_modules.py
+++ b/src/sagemaker_containers/_modules.py
@@ -171,7 +171,7 @@ def download_and_install(uri, name=DEFAULT_MODULE_NAME, cache=True):
             install(module_path)
 
 
-def run(module_name, args=None, env_vars=None, wait=True):  # type: (str, list, dict, bool) -> Popen
+def run(module_name_or_script, args=None, env_vars=None, wait=True, bash=False):  # type: (str, list, dict, bool) -> Popen
     """Run Python module as a script.
 
     Search sys.path for the named module and execute its contents as the __main__ module.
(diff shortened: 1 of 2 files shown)


src/sagemaker_containers/_modules.py
```diff
@@ -171,7 +171,7 @@ def download_and_install(uri, name=DEFAULT_MODULE_NAME, cache=True):
             install(module_path)
 
 
-def run(module_name, args=None, env_vars=None, wait=True):  # type: (str, list, dict, bool) -> Popen
+def run(module_name_or_script, args=None, env_vars=None, wait=True, bash=False):  # type: (str, list, dict, bool) -> Popen
```
jesterhazy: bash is a bad name. script could be any interpreter on the host, controlled by #! line

jesterhazy: also, we don't need this args at all. why not just branch on file extension (py vs sh or py means python anything else means sh)

if that won't work because of what we are doing with module name before calling this, then we could require .sh extension, or refactor the caller so it what it is.

src/sagemaker_containers/_trainer.py
```diff
@@ -41,20 +41,26 @@ def train():
         # TODO: iquintero - add error handling for ImportError to let the user know
         # if the framework module is not defined.
         env = sagemaker_containers.training_env()
+        
+        if env.bash:
```
jesterhazy: flip this test and use env.python_module?

jesterhazy: also, PR looks like it's missing a file because _env.py isn't here

PR #131, opened 2018-11-09
Add nginx template along with port and accept type configurability.

ChoiByungWook:
Added default for region and log level.

Allowed for port to configurable along with accept type.


src/container_support/utils.py
```diff
@@ -47,3 +47,25 @@ def untar_directory(tar_file_path, extract_dir_path):
...
+
+def read_file(path, mode='r'):
+    """Read data from a file.
+     Args:
+        path (str): path to the file.
+        mode (str): mode which the file will be open.
+     Returns:
+    """
+    with open(path, mode) as f:
+        return f.read()
```
orchidmajumder: There is no newline end of the file.

src/container_support/utils.py
```diff
@@ -47,3 +47,25 @@ def untar_directory(tar_file_path, extract_dir_path):
...
+
+def write_file(path, data, mode='w'):
+    """Write data to a file.
+    Args:
+        path (str): path to the file.
+        data (str): data to be written to the file.
+        mode (str): mode which the file will be open.
+    """
+    with open(path, mode) as f:
+        f.write(data)
```
orchidmajumder: nit: both these and the read might not need a new function.

test/test_environment.py
```diff
@@ -130,17 +130,39 @@ def test_model_server_workers(hosting):
...
+
+def test_port(hosting):
+    with patch.dict('os.environ', {'SAGEMAKER_BIND_TO_PORT': '42',
+                                   'SAGEMAKER_REGION': 'us-west-2'}):
+        env = HostingEnvironment(hosting)
+        assert env.port == '42'
+
+
+def test_safe_port_range(hosting):
+    with patch.dict('os.environ', {'SAGEMAKER_SAFE_PORT_RANGE': '1234',
```
orchidmajumder: This is going to be one string like this?

jesterhazy: no it will always be range 1234-5678

src/container_support/environment.py
```diff
@@ -306,14 +311,22 @@ def __init__(self, base_dir=ContainerEnvironment.BASE_DIRECTORY):
         self.use_nginx = os.environ.get(ContainerEnvironment.USE_NGINX_ENV, 'true') == 'true'
         "Use nginx as front-end HTTP server instead of gunicorn."
 
+        self.port = os.environ.get(ContainerEnvironment.SAGEMAKER_BIND_TO_PORT_ENV, '8080')
```
jesterhazy: better name: bind_port or http_port

src/container_support/serving.py
```diff
@@ -134,6 +139,18 @@ def _download_user_module_internal(cls, env):
...
+    def _create_nginx_config(serving_env):
+        template = cs.utils.read_file(nginx_config_template_file)
+        pattern = re.compile(r'%(\w+)%')
+
+        template_values = {
+            'NGINX_HTTP_PORT': serving_env.port
+        }
+        config = pattern.sub(lambda x: template_values[x.group(1)], template)
+        logger.info('nginx config: \n%s\n', config)
+        cs.utils.write_file(nginx_config_file, config)
```
jesterhazy: this path is inside package resources... we should not be modifying things inside our running package.
put the new conf file. use /etc/nginx/nginx.conf (overwriting) or /etc/nginx.conf instead.

ChoiByungWook: Ahhh... gotcha, don't modify the package itself. Makes a lot of sense. I'll use /etc/nginx.conf, should we have some sort of sagemaker identifier in the directory? like /etc/sagemaker/nginx.conf?

ChoiByungWook: Spoke offline, using /etc/sagemaker-nginx.conf

src/container_support/utils.py
ChoiByungWook: spoke offline, going to keep the functions due to convention here: https://github.com/aws/sagemaker-containers/blob/master/src/sagemaker_containers/_files.py#L66

PR #134, opened 2018-11-15
bump to 2.3.0 and update changelog

ChoiByungWook:


CHANGELOG.rst
```diff
@@ -2,6 +2,11 @@
 CHANGELOG
 =========
 
+2.3.0
+=====
+* feature: Allow for dynamic nginx.conf creation
```
laurenyu: nit: I won't hold you to it, but it'd be nice to have a newlines after ll. 6 & 11 to be consistent with the other entries.

PR #136, opened 2018-11-19
Add intermediate output functionality.

nadiaya:
Add intermediate output functionality:
As soon as a user is done with a file under `/opt/ml/output/intermediate` we would get notified by using inotify. We would copy this file under `/opt/ml/output/intermediate/.tmp.sagemaker_s3_sync` folder preserving the same folder structure to prevent it from being further modified. As we copy the file we would add timestamp with microseconds precision to avoid modification during s3 upload. After that we copy the file to s3. We keep the queue of the files we need to move as FIFO.


src/sagemaker_containers/_intermediate_output.py
```diff
@@ -0,0 +1,165 @@
...
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from __future__ import absolute_import
+
+from concurrent.futures import ThreadPoolExecutor
```
mvsusp: I don't think this module is available in Py27

src/sagemaker_containers/_intermediate_output.py
```diff
@@ -0,0 +1,165 @@
...
+from inotify_simple import INotify, flags
```
mvsusp: Import modules, not classes and functions https://github.com/google/styleguide/blob/gh-pages/pyguide.md#224-decision

That can give you collision and naming problems in the future. Mocking is messy, for example.

src/sagemaker_containers/_intermediate_output.py
```diff
@@ -0,0 +1,165 @@
...
+from multiprocessing import Process
+import os
+import shutil
+from six.moves.urllib.parse import urlparse
+import time
+
+import boto3
+from boto3.s3.transfer import S3Transfer
```
mvsusp: same here.

src/sagemaker_containers/_intermediate_output.py
```diff
@@ -0,0 +1,165 @@
...
+                    _copy_file(executor, s3_uploader, watchers[event.wd], event.name)
+
+        last_pass_done = stop_file_exists
+        stop_file_exists = os.path.exists(success_file_path) or os.path.exists(failure_file_path)
+
+    # wait for all the s3 upload tasks to finish and shutdown the executor
+    executor.shutdown(wait=True)
+
+
+def start_intermediate_folder_sync(s3_output_location, region):
```
mvsusp: When you name a function in Python, always think about module + fn because nobody should import just your fn name.

For example, this one fullname is **_intermediate_output. start_intermediate_folder_sync**

src/sagemaker_containers/_intermediate_output.py
```diff
@@ -0,0 +1,165 @@
...
+                # Upload files to s3 if there any files.
+                # There is a potential race condition if upload the file and the see a notification
+                # for it which should cause any problems because when we copy files to temp dir
+                # we add a unique timestamp up to microseconds.
+                if flag is flags.ISDIR:
+                    for folder, dirs, files in os.walk(os.path.join(intermediate_path, event.name)):
+                        wd = inotify.add_watch(folder, watch_flags)
+                        relative_path = os.path.relpath(folder, intermediate_path)
+                        watchers[wd] = relative_path
+                        os.makedirs(os.path.join(tmp_dir_path, relative_path))
```
mvsusp: Does this line break if he folders exist already?

src/sagemaker_containers/_intermediate_output.py
```diff
@@ -0,0 +1,165 @@
...
+    Args:
+        s3_output_location (str): name of the script or module.
+        region (str): the location of the module.
+
+    Returns:
+        (Process): the intermediate output sync daemonic process.
+    """
+    if not s3_output_location or os.path.exists(intermediate_path):
+        logger.debug('Could not initialize intermediate folder sync to s3.')
+        return None
```
mvsusp: return and return None are the same

src/sagemaker_containers/_intermediate_output.py
```diff
@@ -0,0 +1,165 @@
...
+    # create intermediate and intermediate_tmp directories
+    os.makedirs(intermediate_path)
+    os.makedirs(tmp_dir_path)
+
+    # configure unique s3 output location similar to how SageMaker platform does it
+    # or link it to the local output directory
+    url = urlparse(s3_output_location)
+    if url.scheme == 'file':
+        logger.debug('Local directory is used for output. No need to sync any intermediate output.')
+        return
```
mvsusp: nit: I personally avoid returning in the middle of the fn like that, it makes easier to have logic errors

src/sagemaker_containers/_intermediate_output.py
nadiaya: Yes, good point, they could have been created and then deleted and then created again..

src/sagemaker_containers/_intermediate_output.py
nadiaya: I checked and it is, not sure why the official documentation is missing it.

src/sagemaker_containers/_env.py
```diff
@@ -588,6 +601,14 @@ def additional_framework_parameters(self):  # type: () -> dict
         """
         return self._additional_framework_parameters
 
+    def sagemaker_s3_output(self):  # type: () -> str
+        """s3 output directory location provided by the user.
```
jesterhazy: S3 in docstring

src/sagemaker_containers/_env.py
```diff
@@ -588,6 +601,14 @@ def additional_framework_parameters(self):  # type: () -> dict
...
+
+        Returns:
+            str: s3 location uri.
```
jesterhazy: S3. also uri or location don't need both

src/sagemaker_containers/_env.py
jesterhazy: should this be a property?

src/sagemaker_containers/_env.py
```diff
@@ -757,6 +779,16 @@ def output_data_dir(self):  # type: () -> str
         """
         return self._output_data_dir
 
+    @property
+    def output_intermediate_dir(self):  # type: () -> str
+        """The dir to write intermediate output artifacts for them to be synced to S3.
```
jesterhazy: The directory for intermediate outputs artifacts that should be be synced to S3.

src/sagemaker_containers/_env.py
```diff
@@ -757,6 +779,16 @@ def output_data_dir(self):  # type: () -> str
...
+        The directory special behavior is to move artifacts from the training instance to
```
jesterhazy: Wording is confusing. Maybe:

"Any files written to this directory will be uploaded to S3 by a background process while training is in progress."

src/sagemaker_containers/_intermediate_output.py
```diff
@@ -0,0 +1,177 @@
...
+    return time.strftime("%Y-%m-%d-%H-%M-%S-{}".format(moment_us), time.gmtime(moment))
+
+
+def _upload_to_s3(s3_uploader, relative_path, file_path, filename):
+    try:
+        key = os.path.join(s3_uploader['key_prefix'], relative_path, filename)
+        s3_uploader['transfer'].upload_file(file_path, s3_uploader['bucket'], key)
+
+        # delete the original file
+        os.remove(file_path)
```
jesterhazy: is this always necessary or safe? what if someone wants checkpoints synced to s3, but needs to refer to them on later training iterations?

src/sagemaker_containers/_intermediate_output.py
```diff
@@ -0,0 +1,177 @@
...
+    except FileNotFoundError:
+        # Broken link or deleted
+        pass
+    except Exception:
+        logger.exception('Failed to upload file to s3.')
+
+        # delete the original file
+        os.remove(file_path)
```
jesterhazy: this is in try and except branch. move to finally (with exists check to avoid another FNF

src/sagemaker_containers/_intermediate_output.py
jesterhazy: +1

src/sagemaker_containers/_intermediate_output.py
jesterhazy: +1 this is a guard and doesn't depend on the preceding ops so move it up

src/sagemaker_containers/_params.py
```diff
@@ -39,4 +41,4 @@
 FRAMEWORK_TRAINING_MODULE_ENV = 'SAGEMAKER_TRAINING_MODULE'  # type: str
 SAGEMAKER_HYPERPARAMETERS = (
     USER_PROGRAM_PARAM, SUBMIT_DIR_PARAM, ENABLE_METRICS_PARAM, REGION_NAME_PARAM, LOG_LEVEL_PARAM, JOB_NAME_PARAM,
-    DEFAULT_MODULE_NAME_PARAM, TUNING_METRIC_PARAM)  # type: set
+    DEFAULT_MODULE_NAME_PARAM, TUNING_METRIC_PARAM, S3_OUTPUT_LOCATION_PARAM)  # type: set
```
jesterhazy: type annotation -- not new, but wrong b/c this is a tuple

src/sagemaker_containers/_trainer.py
```diff
@@ -37,11 +38,19 @@ def _exit_processes(exit_code):  # type:
 
 
 def train():
+    intemediate_sync = None
```
jesterhazy: spelling

src/sagemaker_containers/_trainer.py
```diff
@@ -37,11 +38,19 @@ def _exit_processes(exit_code):  # type:
...
     try:
         # TODO: iquintero - add error handling for ImportError to let the user know
         # if the framework module is not defined.
         env = sagemaker_containers.training_env()
 
+        # TODO: There is a bug in the logic - we need os.environ.get(_params.REGION_NAME_ENV)
```
jesterhazy: add owner + fix eta or GH issue

src/sagemaker_containers/_trainer.py
```diff
@@ -59,7 +68,12 @@ def train():
             entry_point.run(env.module_dir, env.user_entry_point, env.to_cmd_args(), env.to_env_vars())
 
         logger.info('Reporting training SUCCESS')
+
         files.write_success_file()
+
+        if intemediate_sync:
```
jesterhazy: in multiple branches. can we consolidate in one finally?

src/sagemaker_containers/_env.py
nadiaya: No, i don't want it to show up in the customer logs as available training env property.
I assume we will remove it when platform supports it.

src/sagemaker_containers/_intermediate_output.py
nadiaya: we are removing our own temporarily files that we created, e.g.
- user writes intermediate/file1.txt -> we create intermediate/.tmp_s3_intermediate/2018-11-28:19:30_file1.txt 
- user updates intermediate/file1.txt -> we create intermediate/.tmp_s3_intermediate/2018-11-28:19:31_file1.txt  (with a different timestamp)
- upload intermediate/.tmp_s3_intermediate/2018-11-28:19:30_file1.txt to s3 as file1.txt, delete intermediate/.tmp_s3_intermediate/2018-11-28:19:30_file1.txt
- upload intermediate/.tmp_s3_intermediate/2018-11-28:19:31_file1.txt to s3 as file1.txt (overwrites original one), delete intermediate/.tmp_s3_intermediate/2018-11-28:19:31_file1.txt

src/sagemaker_containers/_trainer.py
nadiaya: i don't think so :(
because we need to exist with different exit codes the main process, and we can't do anything after we finished main process...

src/sagemaker_containers/_trainer.py
nadiaya: sure :)

src/sagemaker_containers/_env.py
jesterhazy: ok

src/sagemaker_containers/_intermediate_output.py
jesterhazy: right ok.

src/sagemaker_containers/_trainer.py
jesterhazy: ok

src/sagemaker_containers/_intermediate_output.py
jesterhazy: ok. side effect before the check is wanted even if sync is disabled

src/sagemaker_containers/_env.py
```diff
@@ -781,9 +781,9 @@ def output_data_dir(self):  # type: () -> str
 
     @property
     def output_intermediate_dir(self):  # type: () -> str
-        """The dir to write intermediate output artifacts for them to be synced to S3.
-        The directory special behavior is to move artifacts from the training instance to
-        s3 directory during training in case if sagemaker_s3_output was specified.
+        """The directory for intermediate outputs artifacts that should be synced to S3.
```
jesterhazy: "output artifacts"

PR #141, opened 2018-11-21
Re-introduce public methods that were removed in v2.3.1 and bump version to v2.3.5

laurenyu:
A few things are happening in this PR:
1. add changelog entries to clearly denote the breaking changes we released in v2.3.1
1. re-add ``_modules.s3_download`` and ``_modules.download_and_install``
1. bump version to v2.3.5

### DIFF
--- a/src/sagemaker_containers/_modules.py
+++ b/src/sagemaker_containers/_modules.py
@@ -16,6 +16,7 @@ import importlib
 import os
 import shlex
 import sys
+import tarfile
 import textwrap
 import warnings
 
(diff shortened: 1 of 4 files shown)


src/sagemaker_containers/_modules.py
```diff
@@ -108,6 +109,58 @@ def install(path, capture_error=False):  # type: (str, bool) -> None
...
+    for backward-compatibility purposes.
+
+    Args:
+        url (str): the S3 URL of the file.
+        dst (str): the destination where the file will be saved.
+    """
+    _files.s3_download(url, dst)
+
+
+def download_and_install(uri, name=DEFAULT_MODULE_NAME, cache=True):
```
mvsusp:
```python
def download_and_install(uri, name=DEFAULT_MODULE_NAME, cache=True):
     _files.download_and_extract(name, uri, _env.code_dir)
     prepare(_env.code_dir, name)
     install(_env.code_dir)
```

laurenyu: are there any potential side effects of permanently saving it in `code_dir` and using that location for caching vs saving the file in wherever it gets pip installed?

mvsusp: _modules.install still installs the module using pip, the functionality was untouched https://github.com/aws/sagemaker-containers/blob/master/src/sagemaker_containers/_modules.py#L101

The file will in the pip location as well. Do you see any issues for copying the file to code dir instead of a source dir?

laurenyu: I can't think of anything off the top of my head, but I also haven't looked thoroughly enough to say with certainty that changing where the file is copied to won't make a difference. Given that this PR is about preserving previous behavior, I'd rather not introduce another change if we're not sure.

mvsusp: If you want to preserve behavior, although I don't see any issues in the suggested approach you can do the following as well:

```python
def download_and_install(uri, name=DEFAULT_MODULE_NAME, cache=True):
    should_use_cache = cache and exists(name)
     if not should_use_cache:
        with _files.tmpdir() as tmpdir:
        _files.download_and_extract(name, uri, tmpdir)
        prepare(tmpdir, name)
        install(tmpdir)
```

That being said, I am okay with your current implementation, besides the repeated code.

laurenyu: I'm good with your second suggestion there since it preserves behavior - definitely agree about avoiding repeating code. I'll push a change with that once I get the functional tests working

PR #142, opened 2018-11-22
bug-fix: Accept header ANY ('*/*') fallback to default accept

mvsusp:
*Issue #, if available:*
bug-fix: Accept header ANY ('/') fallback to default accept

*Description of changes:*
1.0 change: https://github.com/aws/sagemaker-containers/pull/143

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #143, opened 2018-11-23
bug-fix: Accept header ANY ('*/*') fallback to default accept

mvsusp:
*Issue #, if available:*
bug-fix: Accept header ANY ('*/*') fallback to default accept

*Description of changes:*
2.0 change: https://github.com/aws/sagemaker-containers/pull/142

### DIFF
--- a/src/container_support/serving.py
+++ b/src/container_support/serving.py
@@ -207,7 +207,11 @@ class Server(object):
         # Accepting both ContentType and Content-Type headers. ContentType because Coral and Content-Type because,
         # well, it is just the html standard
         input_content_type = request.headers.get('ContentType', request.headers.get('Content-Type', JSON_CONTENT_TYPE))
+
         requested_output_content_type = request.headers.get('Accept') or env.default_accept or JSON_CONTENT_TYPE
+
+        if requested_output_content_type == ANY_CONTENT_TYPE:
+            requested_output_content_type = env.default_accept
 
         # utf-8 decoding is automatic in Flask if the Content-Type is valid. But that does not happens always.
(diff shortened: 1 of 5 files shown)


ChoiByungWook:
Changed the to merge branch from master to r1.0 since this is container_support changes.

src/container_support/serving.py
```diff
@@ -207,8 +207,12 @@ def _invoke(self):
         # Accepting both ContentType and Content-Type headers. ContentType because Coral and Content-Type because,
         # well, it is just the html standard
         input_content_type = request.headers.get('ContentType', request.headers.get('Content-Type', JSON_CONTENT_TYPE))
+
         requested_output_content_type = request.headers.get('Accept') or env.default_accept or JSON_CONTENT_TYPE
 
+        if requested_output_content_type == ANY_CONTENT_TYPE:
+            requested_output_content_type = env.default_accept
```
ChoiByungWook: I believe this will end up being None if the environment variable isn't set.

https://github.com/aws/sagemaker-containers/blob/r1.0/src/container_support/environment.py#L318

I believe if the user uses boto3 and doesn't specify an accept type, then boto3 will default to */*.

test/test_serving.py
```diff
@@ -57,6 +57,40 @@ def test_app_invoke(app):
     assert "application/json" == result.content_type
 
 
+def test_app_invoke_with_accept_any():
+    with patch.dict('os.environ', {'SAGEMAKER_DEFAULT_INVOCATIONS_ACCEPT': 'application/NPY'}):
```
jesterhazy: nit: should to use something that looks like valid mimetype even though test doesn't care (e.g. application/x-npy)

PR #153, opened 2018-12-14
Generic MPI - Create library to changehostname when mpi starts

mvsusp:
*Description of changes:*
- Create libchangehostname library necessary to run mpi in SageMaker


.flake8
```diff
@@ -1,3 +1,3 @@
 [flake8]
-application_import_names = sagemaker_containers, test
+application_import_names = sagemaker_containers, test, libchangehostname
 import-order-style = google
```
laurenyu: nit: newline at the end

src/sagemaker_containers/c/libchangehostname.c
```diff
@@ -0,0 +1,68 @@
...
+    return PyModule_Create(&libchangehostnamemodule);
+}
+#else
+PyMODINIT_FUNC initlibchangehostname() {
+    PyObject* module;
+
+    module = Py_InitModule3(
+        "libchangehostname", LibchangehostnameMethods, "Returns the value of $SM_CURRENT_HOST");
+}
+#endif
```
laurenyu: nit: newline at the end of the file

src/sagemaker_containers/c/libchangehostname.c
```diff
@@ -0,0 +1,68 @@
...
+// ANY KIND, either express or implied. See the License for the specific
+// language governing permissions and limitations under the License.
+
+#include <Python.h>
+#include <stdlib.h>
+
+
+int libchangehostname(char *name, size_t len)
+{
+  const char *val = getenv("SM_CURRENT_HOST");
```
nadiaya: i don't really like that it depends on an environment variable that might now be initialized yet.
I understand that right now we call it at 'the right' time, but since it's not really guaranteed by anyone I am worried that it might cause issues in the future.
Also sagemaker-containers is not just a library used by us, we also advertise and help customers to use it in BYOC use case.

Anyway we can switch to use json file or guarantee that it's called only after this environment variable is initialized?

PR #157, opened 2018-12-19
Generic OpenMPI support

mvsusp:
*Description of changes:*
- generic OpenMPI support

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ packages.append('sagemaker_containers.etc')
 
 required_packages = [
     'numpy', 'boto3', 'six', 'pip', 'flask', 'gunicorn', 'typing',
-    'gevent', 'inotify_simple', 'werkzeug', 'paramiko'
+    'gevent', 'inotify_simple', 'werkzeug', 'paramiko==2.4.2', 'psutil==5.4.8'
 ]
 
 # enum is introduced in Python 3.4. Installing enum back port
--- a/setup.py
+++ b/setup.py
@@ -16,25 +16,31 @@ from glob import glob
 import os
 import sys
 
-from setuptools import find_packages, setup
+import setuptools
 
 
 def read(file_name):
     return open(os.path.join(os.path.dirname(__file__), file_name)).read()
 
 
-packages = find_packages(where='src', exclude=('test',))
+packages = setuptools.find_packages(where='src', exclude=('test',))
 packages.append('sagemaker_containers.etc')
 
 required_packages = [
-  'boto3', 'six', 'pip', 'flask', 'gunicorn', 'gevent', 'inotify_simple', 'werkzeug'
+    'numpy', 'boto3', 'six', 'pip', 'flask', 'gunicorn', 'typing',
+    'gevent', 'inotify_simple', 'werkzeug', 'paramiko'
 ]
 
 # enum is introduced in Python 3.4. Installing enum back port
 if sys.version_info < (3, 4):
     required_packages.append('enum34 >= 1.1.6')
 
-setup(
+gethostname = setuptools.Extension('libchangehostname',
+                                   sources=['src/sagemaker_containers/c/libchangehostname.c'],
+                                   extra_compile_args=['-Wall', '-shared', '-export-dynamic',
+                                                       '-ldl'])
+
+setuptools.setup(
     name='sagemaker_containers',
     version='2.3.5',
     description='Open source library for creating containers to run on Amazon SageMaker.',
@@ -46,6 +52,7 @@ setup(
     },
     package_data={'sagemaker_containers.etc': ['*']},
     py_modules=[os.path.splitext(os.path.basename(path))[0] for path in glob('src/*.py')],
+    ext_modules=[gethostname],
     long_description=read('README.md'),
     author='Amazon Web Services',
     url='https://github.com/aws/sagemaker-containers/',
@@ -64,11 +71,11 @@ setup(
     install_requires=required_packages,
 
     extras_require={
-        'test': ['tox', 'flake8', 'pytest', 'pytest-cov', 'mock', 'sagemaker', 'numpy']
+        'test': ['tox', 'flake8', 'pytest', 'pytest-cov', 'mock', 'sagemaker==1.16.2']
     },
 
     entry_points={
-          'console_scripts': ['serve=sagemaker_containers.cli.serve:main',
-                              'train=sagemaker_containers.cli.train:main'],
+        'console_scripts': ['serve=sagemaker_containers.cli.serve:main',
+                            'train=sagemaker_containers.cli.train:main'],
     }
 )
--- a/src/sagemaker_containers/_env.py
+++ b/src/sagemaker_containers/_env.py
@@ -138,7 +138,6 @@ Returns:
     str: the path to the intermediate output directory, e.g. /opt/ml/output/intermediate.
 """
 
-
 HYPERPARAMETERS_FILE = 'hyperparameters.json'  # type: str
 RESOURCE_CONFIG_FILE = 'resourceconfig.json'  # type: str
 INPUT_DATA_CONFIG_FILE = 'inputdataconfig.json'  # type: str
@@ -164,7 +163,7 @@ def _create_training_directories():
 
     resources_dict = {
         "current_host": host_name,
-        "hosts":        [host_name]
+        "hosts": [host_name]
     }
     _write_json(resources_dict, resource_config_file_dir)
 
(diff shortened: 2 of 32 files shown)


src/sagemaker_containers/_mpi.py
```diff
@@ -0,0 +1,236 @@
...
+
+3. Create SSH key to allow password less ssh between different docker instances:
+mkdir -p /root/.ssh/ && ssh-keygen -q -t rsa -N '' -f /root/.ssh/id_rsa && \
+  cp /root/.ssh/id_rsa.pub /root/.ssh/authorized_keys && \
+  printf "Host *\n  StrictHostKeyChecking no\n" >> /root/.ssh/config
+"""
+
+
+def _start_sshd_daemon():  # type: () -> None
+    sshd_executable = '/usr/sbin/sshd'
```
andremoeller: What if sshd is at /usr/local/sbin/sshd instead? Then this will erroneously fail.

I think it would be preferable to try to run sshd and raise a RuntimeError with this message if sshd isn't anywhere on the PATH.

src/sagemaker_containers/_mpi.py
```diff
@@ -0,0 +1,236 @@
...
+
+    def __init__(self, user_entry_point, args, env_vars, master_hostname, hosts, process_per_host,
+                 custom_mpi_options, network_interface_name, interval=1,
+                 timeout_in_seconds=60 * 60):
+
+        super(MasterRunner, self).__init__(user_entry_point, args, env_vars)
+
+        self._master_hostname = master_hostname
+        self._hosts = hosts
+        self._process_per_host = process_per_host
```
andremoeller: "process_per_host" -> "num_processes_per_host" or "processes_per_host"

src/sagemaker_containers/_mpi.py
```diff
@@ -0,0 +1,236 @@
...
+    for i in range(5 * 60):
+        procs = [p for p in psutil.process_iter(attrs=['name']) if p.info['name'] == 'orted']
+
+        if procs:
+            return procs
+        time.sleep(1)
+
+
+class MasterRunner(_process.ProcessRunner):
+    """Responsible to prepare MPI distributed training and syncronize work with the Workers.
```
andremoeller: "Responsible to prepare MPI distributed training and syncronize work with the Workers"
->
"Responsible for preparing MPI distributed training and synchronizing work with the Workers"

src/sagemaker_containers/_mpi.py
```diff
@@ -0,0 +1,236 @@
...
+
+import libchangehostname
+from sagemaker_containers import _logging, _process, _timeout
+
+logger = _logging.get_logger()
+logging.getLogger("paramiko").setLevel(logging.INFO)
+
+
+class WorkerRunner(_process.ProcessRunner):
+    """Runner responsible prepare MPI training and wait for MPI master execution
```
andremoeller: "Runner responsible prepare MPI training and wait for MPI master execution"
->
"Runner responsible for preparing MPI distributed training and waiting for MPI master execution to finish."

src/sagemaker_containers/_mpi.py
```diff
@@ -0,0 +1,236 @@
...
+
+    def __init__(self, user_entry_point, args, env_vars, master_hostname):
+        super(WorkerRunner, self).__init__(user_entry_point, args, env_vars)
+        self._master_hostname = str(master_hostname)
+
+    def run(self, wait=True, capture_error=False):  # type: (bool, bool) -> None
+        """The WorkerRunner proceeds as following:
+
+        - wait for the MPI Master to create its SSH daemon
+        - start its SSH daemon
```
andremoeller: I'd reorder and reword these:

- Start the worker's SSH daemon
- Wait for the MPI Master to create its own SSH daemon.
...

src/sagemaker_containers/_mpi.py
```diff
@@ -0,0 +1,236 @@
...
+            time.sleep(1)
+
+    def _wait_master_to_finish(self):  # type: () -> None
+        while _can_connect(self._master_hostname):
+            time.sleep(30)
+
+
+def _wait_orted_process_to_finish():  # type: () -> None
+    orted = _orted_process()
+    psutil.wait_procs(orted)
```
andremoeller: What if orted is empty (say, if the orted process doesn't start in time)? Will psutil.wait_process exit immediately?

The job should fail in this case -- I recommend a timeout below.

src/sagemaker_containers/_mpi.py
```diff
@@ -0,0 +1,236 @@
...
+
+
+def _orted_process():
+    """Waits maximum of 5 minutes for orted process to start"""
```
andremoeller: I think we should use _timeout.timeout here as well.

src/sagemaker_containers/_mpi.py
```diff
@@ -0,0 +1,236 @@
...
+
+        _start_sshd_daemon()
+
+        if wait:
+            logger.info('Waiting for MPI process to finish.')
+            _wait_orted_process_to_finish()
+            time.sleep(30)
+        logger.info('MPI process finished.')
+
+    def _wait_master_to_start(self):  # type: () -> None
```
andremoeller: I think this should have a timeout (using _timeout.timeout) in case the worker can never connect to the master.

src/sagemaker_containers/_mpi.py
```diff
@@ -0,0 +1,236 @@
...
+        time.sleep(1)
+
+
+class MasterRunner(_process.ProcessRunner):
+    """Responsible to prepare MPI distributed training and syncronize work with the Workers.
+    """
+
+    def __init__(self, user_entry_point, args, env_vars, master_hostname, hosts, process_per_host,
+                 custom_mpi_options, network_interface_name, interval=1,
+                 timeout_in_seconds=60 * 60):
```
andremoeller: An hour seems huge for this timeout. For Chainer, workers would come up within seconds of each other, even for large clusters.

Could this be 1 minute? A shorter timeout benefits users since their jobs will fail faster (and cheaper).

test/resources/openmpi/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the 'license' file accompanying this file. This file is
+# distributed on an 'AS IS' BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+FROM mvsusp/openmpi
```
andremoeller: Is this the Docker image built from Dockerfile.base? If so, could you document this?

Also: seems odd to have it under a personal namespace -- I think this should either be put in a team-owned repository or assumed that the base is built locally with a certain name.

setup.py
```diff
@@ -16,25 +16,31 @@
...
+    'numpy', 'boto3', 'six', 'pip', 'flask', 'gunicorn', 'typing',
+    'gevent', 'inotify_simple', 'werkzeug', 'paramiko==2.4.2', 'psutil==5.4.8'
 ]
 
 # enum is introduced in Python 3.4. Installing enum back port
 if sys.version_info < (3, 4):
     required_packages.append('enum34 >= 1.1.6')
 
-setup(
+gethostname = setuptools.Extension('libchangehostname',
```
andremoeller: Nice!

setup.py
```diff
@@ -64,11 +71,11 @@ def read(file_name):
     install_requires=required_packages,
 
     extras_require={
-        'test': ['tox', 'flake8', 'pytest', 'pytest-cov', 'mock', 'sagemaker', 'numpy']
+        'test': ['tox', 'flake8', 'pytest', 'pytest-cov', 'mock', 'sagemaker==1.16.2']
```
andremoeller: ==1.16.2, or >=1.16.2? If a breaking change is released to SageMaker Python SDK, the tests should fail.

src/sagemaker_containers/_env.py
```diff
@@ -547,14 +546,17 @@ def __init__(self, resource_config=None, input_data_config=None, hyperparameters
...
+                                                         boto3.session.Session().region_name)
 
         os.environ[_params.JOB_NAME_ENV] = sagemaker_hyperparameters.get(_params.JOB_NAME_PARAM, '')
         os.environ[_params.CURRENT_HOST_ENV] = current_host
         os.environ[_params.REGION_NAME_ENV] = sagemaker_region or ''
 
         self._hosts = hosts
-        self._network_interface_name = network_interface_name
+
+        self._network_interface_name = resource_config.get('network_interface_name', 'eth0')
```
andremoeller: Why 'eth0' and not 'ethwe'? (In fact, why either at all? Could you add a comment?)

src/sagemaker_containers/_mpi.py
```diff
@@ -0,0 +1,236 @@
...
+                   '--host', ','.join(host_list),
+                   '-np', str(num_processes),
+
+                   '--allow-run-as-root',
+                   '--display-map',
+                   '--tag-output',
+
+                   '-mca', 'btl_tcp_if_include', self._network_interface_name,
+                   '-mca', 'oob_tcp_if_include', self._network_interface_name,
+                   '-mca', 'plm_rsh_no_tree_spawn', '1',
```
andremoeller: Are all of these necessary? I think we should bias towards not including options rather than including options that may not be necessary.

A few suspect ones are plm_rsh_no_tree_spawn, NCCL_DEBUG, and NCCL_MIN_NRINGS.

andremoeller:
@mvsusp 

To clarify: we talked offline, but I wouldn't ship this if it were up to me.

PR #161, opened 2019-01-09
Re add skipped test

mvsusp:


test/resources/openmpi/Dockerfile
```diff
@@ -10,14 +10,10 @@
 # distributed on an 'AS IS' BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
-FROM mvsusp/openmpi
+FROM 142577830533.dkr.ecr.us-west-2.amazonaws.com/openmpi
```
ChoiByungWook: Can we use a public image from Dockerhub?

I understand this is a test, however I believe this limits the ability for others to test to only the SageMaker development team.

ChoiByungWook: I imagine using the FROM solution is cleaner, however I feel that this abstracts away something that isn't necessarily available publicly or is public without a lot of documentation and confusion.

PR #164, opened 2019-01-21
Fix recursive directory navigation

saurabh3949:
Since event.name of inotify just returns the folder name if a new folder is created, the directory structure wasn't being walked properly. This change fixes the bug to enable correct traversal of the directory.

### DIFF
--- a/src/sagemaker_containers/_intermediate_output.py
+++ b/src/sagemaker_containers/_intermediate_output.py
@@ -97,7 +97,10 @@ def _watch(inotify, watchers, watch_flags, s3_uploader):
                 # for it which should cause any problems because when we copy files to temp dir
                 # we add a unique timestamp up to microseconds.
                 if flag is inotify_simple.flags.ISDIR and inotify_simple.flags.CREATE & event.mask:
-                    for folder, dirs, files in os.walk(os.path.join(intermediate_path, event.name)):
+                    path = os.path.join(intermediate_path,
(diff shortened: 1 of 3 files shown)


src/sagemaker_containers/_intermediate_output.py
```diff
@@ -97,7 +97,9 @@ def _watch(inotify, watchers, watch_flags, s3_uploader):
                 # for it which should cause any problems because when we copy files to temp dir
                 # we add a unique timestamp up to microseconds.
                 if flag is inotify_simple.flags.ISDIR and inotify_simple.flags.CREATE & event.mask:
-                    for folder, dirs, files in os.walk(os.path.join(intermediate_path, event.name)):
+                    path = os.path.join(
+                    *[intermediate_path] + [watchers[wd] for wd in range(1, event.wd + 1)] + [event.name])
```
mvsusp: Please extract this logic outside `os.path.join` to be more readable and easier to understand.

saurabh3949: There was an issue with this. I have fixed it in the latest commit.

nadiaya:
We have a test that was supposed to test for this behavior, but failed :(
We should add test that creates dir1 and dir1/dir2 with a sleep in between them.

nadiaya:
Tests failed because of this numpy bug:
https://github.com/numpy/numpy/pull/12754

PR #168, opened 2019-02-05
Only load environment variables once

icywang86rui:
Currently we are loading all the hosting related environment varialbes on every
request. This change should decrease inference latency.

*Issue #, if available:*

*Description of changes:*


icywang86rui:
I will add unit testing. We don't have any long running inference tests that can give us metrics. I think we should have one. Had a brief conversation about this with Jonathon about this.

PR #171, opened 2019-02-15
Remove numpy restriction.

nadiaya:
Remove numpy restriction.


ChoiByungWook:
Why are we removing this numpy restriction?

nadiaya:
The restrictions was added because of a numpy bug introduced in 16.0. It looks like it fixed in 16.1 (our functional tests would fail because of it if it's not fixed). Having a numpy version fixed causes version conflicts with latest TF and coach versions.

src/sagemaker_containers/_mapping.py
```diff
@@ -93,7 +93,7 @@ def arg_value(value):
     return [item for item in itertools.chain.from_iterable(items)]
 
 
-def _decode(obj):  # type: (bytes or str or unicode or object) -> unicode
+def _decode(obj):  # type: (bytes or str or unicode or object) -> unicode # noqa ignore=F821
```
laurenyu: I'd remove `unicode` like with above. it's only valid in Python 2, which doesn't check these annotations

src/sagemaker_containers/_intermediate_output.py
```diff
@@ -61,7 +61,7 @@ def _copy_file(executor, s3_uploader, relative_path, filename):
         dst = os.path.join(tmp_dir_path, relative_path, '{}.{}'.format(_timestamp(), filename))
         shutil.copy2(src, dst)
         executor.submit(_upload_to_s3, s3_uploader, relative_path, dst, filename)
-    except FileNotFoundError:
+    except FileNotFoundError:  # noqa ignore=F821
```
laurenyu: I think it's nicer to configure flake8 to understand this is a built-in error: https://github.com/aws/sagemaker-containers/pull/170/files#diff-b91f3d5bd63fcd17221b267e851608e8R20

src/sagemaker_containers/_mapping.py
nadiaya: i wanted to keep this function mentioning it since it's explicitly differentiate  between python2 unicode and python3 strings, for humans, not for flake8:)

laurenyu: gotcha, fair enough

PR #176, opened 2019-02-21
Switch README to RST

laurenyu:
*Description of changes:*
The README isn't displaying correctly on PyPI, and since all of our other READMEs are RST rather than markdown, this seems like the quicker fix.

NB: I made changes only for formatting. I intentionally did not make any content changes because I figured that'd get obscured in the diff and make it hard to review.


jesterhazy:
maybe this is failing because we have out of date versions of setuptools and/or twine on the build agent?

laurenyu:
@jesterhazy the build errors were things like `UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 13662: ordinal not in range(128)`, which we've definitely seen before because of characters like fancy quotes or non-breaking spaces

PR #232, opened 2019-10-22
Use pandas in csv encoding/decoding for correct quoted string handling.

aws-patlin:
*Description of changes:*
CSV encoding and decoding for inference does not have the correct behavior when handling string fields that embed special characters (i.e. commas, newlines, double quotes). The expected behavior for the text/csv MIME type can be found in [this document](https://tools.ietf.org/html/rfc4180#section-2). I've replaced the numpy logic in `_encoders.csv_to_numpy` and `_encoders.array_to_csv` with the built-in csv library, which is able to properly read and write in this format.

Additional unit tests were added to test handling of these special characters. Tests run with tox and passing, excluding test_mpi.py (skipped), test_gethostname.py (xfailed), and pylint errors on _mapping.py which already occur before this change.

## Merge Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your pull request._

- [x] I have read the [CONTRIBUTING](https://github.com/aws/sagemaker-containers/blob/master/CONTRIBUTING.md) doc
- [x] I used the commit message format described in [CONTRIBUTING](https://github.com/aws/sagemaker-containers/blob/master/CONTRIBUTING.md#commit-message-guidlines)
- [x] I have added tests that prove my fix is effective or that my feature works (if appropriate)
- [x] I have updated any necessary [documentation](https://github.com/aws/sagemaker-containers/blob/master/README.rst) (if appropriate)

### DIFF
--- a/src/sagemaker_containers/_encoders.py
+++ b/src/sagemaker_containers/_encoders.py
@@ -13,6 +13,7 @@
 """Placeholder docstring"""
 from __future__ import absolute_import
 
+import csv
 import json
 from typing import Iterable
 
@@ -104,8 +105,19 @@ def csv_to_numpy(string_like, dtype=None):  # type: (str) -> np.array
     Returns:
         (np.array): numpy array
     """
-    stream = StringIO(string_like)
-    return np.genfromtxt(stream, dtype=dtype, delimiter=",")
+    try:
+        stream = StringIO(string_like)
+        reader = csv.reader(stream, delimiter=",", quotechar='"', doublequote=True, strict=True)
+        array = np.array([row for row in reader]).squeeze()
+        array = array.astype(dtype)
+    except ValueError as e:
+        if dtype is not None:
+            raise _errors.ClientError(
+                "Error while writing numpy array: {}. dtype is: {}".format(e, dtype)
+            )
+    except Exception as e:
+        raise _errors.ClientError("Error while decoding csv: {}".format(e))
+    return array
 
 
 def array_to_csv(array_like):  # type: (np.array or Iterable or int or float) -> str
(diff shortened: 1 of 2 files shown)


aws-patlin:
There seems to be an error during the CodeBuild due to black-check. Can I get some context on what this is and how to fix the issue?

knakad:
tl;dr: Run the following locally and repush =)
```
tox tests/unit
```

The log states:
"would reformat /codebuild/output/src633343977/src/github.com/aws/sagemaker-containers/test/unit/test_encoder.py
All done! \U0001f4a5 \U0001f494 \U0001f4a5
1 file would be reformatted, 66 files would be left unchanged.
ERROR: InvocationError for command /codebuild/output/src633343977/src/github.com/aws/sagemaker-containers/.tox/black-check/bin/black -l 100 --check ./ (exited with code 1)
"

This essentially means that test_encoder.py is not formatted as black would like. I recommend running black. It should automatically run as part of the following command. I recommend running the following command to run all the unit tests as well as the black formatting tool:
```
tox tests/unit --parallel all
```

If you need to run black on its own, you can also run the black command manually:
https://github.com/aws/sagemaker-containers/blob/master/tox.ini#L82 :
```
black -l 100 ./
```

Hope that helps!

aws-patlin:
I'll take a look to see what other options there are, but pandas is definitely one of the most robust libraries for this.

aws-patlin:
I looked into the built-in python `csv` library, which should be sufficient for this problem. I will push a new commit with the updated logic.

This introduces an false positive error in pylint due to numpy: `src/sagemaker_containers/_encoders.py:137:35: E1136: Value 'array.shape' is unsubscriptable (unsubscriptable-object)`. I haven't figured out how to resolve this.

laurenyu:
sagemaker-containers is only 49KB in size (according to PyPI), so 10 MB would make a difference. (also the teams who build Docker images are usually striving to make their images as small as possible.)

for the pylint error, it looks like it's a known bug: https://github.com/PyCQA/pylint/issues/2885. Could you just add a `# pylint: disable=unscriptable-object` in the line of code that's causing it?

laurenyu:
thanks for doing this!

aws-patlin:
No problem. Likewise, thanks for your help!

PR #244, opened 2020-02-05
added split path function and tests, inserted into _process.py

loodvn:
*Issue #, if available:*
#245

*Description of changes:*

## Merge Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your pull request._

- [x] I have read the [CONTRIBUTING](https://github.com/aws/sagemaker-containers/blob/master/CONTRIBUTING.md) doc
- [ ] I used the commit message format described in [CONTRIBUTING](https://github.com/aws/sagemaker-containers/blob/master/CONTRIBUTING.md#commit-message-guidlines)
- [x] I have added tests that prove my fix is effective or that my feature works (if appropriate)
- [x] I have updated any necessary [documentation](https://github.com/aws/sagemaker-containers/blob/master/README.rst) (if appropriate)


loodvn:
As described in https://github.com/aws/sagemaker-containers/issues/245, I can't seem to use an entry point that's nested inside my package.

In _process.py, we trim the ".py" extension from e.g. `train.py` and then run 

python -m train <args>

When the user-given entry point contains path seperators, this becomes

python -m my_package/train <args>

, which causes an ImportError.

I have two proposed ideas for a fix:

* Replace path separators with periods
* If "/" exists in the entry point, change the entry point type to "python program", which means that my_package/train.py will run as a program.

I'm just not sure if the second would work, so I implemented the first so long.

loodvn:
Oops, I just realised that with something like 

setup.py
src/
  my_package/
    train.py

The above change would still give me src.my_package.train, instead of my_package.train.

Do you have any thoughts on this?

laurenyu:
going to close this in favor of https://github.com/aws/sagemaker-python-sdk/pull/941 (see #245)

